PR #55, opened 2023-08-23
feat: add support for didcomm credentials

TimoGlastra:
Adds support for receiving and showing anoncreds credentials issued over didcomm. 

Does not include sharing proofs over DIDComm yet (will be handled in a follow up PR).

(see video in slack)


Tommylans (approved):
LGTM

PR #57, opened 2023-08-28
feat: add mediation setup

TimoGlastra:
Correctly setup mediation in the wallet. If no internet connection is availalble you can still open the wallet and view your credentials etc.., but not accept incoming requests.

It now also displays a toast when the network connection is lost, and when it is restored again.

The mediationInvitationUrl will be removed once the mediator is live and we can use a did


apps/expo/app/_layout.tsx
```diff
@@ -57,6 +75,23 @@ export default function HomeLayout() {
     void startAgent()
   }, [])
 
+  // Setup mediation
+  useEffect(() => {
+    if (!agent) return
+    if (!hasInternetConnection || isMediationConfigured) return
+
+    void hasMediationConfigured(agent).then(async (mediationConfigured) => {
```
Tommylans, line 83: Do you think there is a chance of this triggered twice before the async function is actually resolved? Or will that be handled inside afj?

Tommylans (approved):
One question the rest LGTM

apps/expo/app/_layout.tsx
TimoGlastra: Yeah good point. I can add a state param that tracks whether the state is currently loading

PR #58, opened 2023-08-28
feat: add inbox

TimoGlastra:
https://github.com/animo/paradym-wallet/assets/23165168/460f3b4c-a874-49a1-99c6-ba228dad4afa


packages/app/components/NoContentInbox.tsx
```diff
@@ -0,0 +1,22 @@
...
+  const { back } = useRouter()
+
+  return (
+    <YStack jc="space-between" px="$4" height="95%">
+      <Spacer />
+      <YStack>
+        <YStack jc="center" ai="center" gap="$2">
+          <Heading variant="h2">You're all caught up.</Heading>
+          <Paragraph textAlign="center" secondary>
+            You don't have any notifications at the moment.
```
Tommylans, line 14: Maybe this a bit smaller than the title?

Tommylans (approved):
Looks good maybe a small styling improvement but I'm not sure of how it would look. It can also look a bit odd maybe good to ask @janrxyz later for a quick look :smile:.

packages/app/components/NoContentInbox.tsx
TimoGlastra: I kept it the same as the home screen when you don't have any crednetials yet. Shouild i also change that you think for consistency?

TimoGlastra: Merged for now, but let me know, and I'll make a follow up PR

PR #62, opened 2023-09-01
fix: qr scanner now reactivates and shows err msg

karimStekelenburg:
Signed-off-by: Karim Stekelenburg <karim@animo.id>

### DIFF
--- a/packages/agent/src/parsers.ts
+++ b/packages/agent/src/parsers.ts
@@ -75,8 +75,8 @@ export const receiveCredentialFromOpenId4VciOffer = async ({
         supportsAllDidMethods || supportedDidMethods.includes('did:jwk')
           ? 'jwk'
           : supportedDidMethods.includes('did:key')
-            ? 'key'
-            : undefined
+          ? 'key'
+          : undefined
 
       if (!didMethod) {
         throw new Error(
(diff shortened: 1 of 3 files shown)


packages/agent/src/parsers.ts
```diff
@@ -259,12 +261,21 @@ export async function receiveOutOfBandInvitation(
...
+    }))
 
-  // Assign connectionId so it can be used in the observables.
-  connectionId = connectionRecord?.id
+    // Assign connectionId so it can be used in the observables.
+    connectionId = connectionRecord?.id
+  } catch (error) {
+    return {
+      result: 'error',
+      message: 'Invitation has already been scanned.',
```
TimoGlastra, line 276: There are quite some errors that could occur here. Can we somehow catch whether it really is the case that this invitation already has been scanned?

Maybe we can try to find an invitation by the id, and if a record is found we show the notification. That way we know for sure it is actually already scanned (instead of another error)

packages/agent/src/parsers.ts
```diff
@@ -259,12 +261,21 @@ export async function receiveOutOfBandInvitation(
     )
   )
 
-  const { connectionRecord, outOfBandRecord } = await agent.oob.receiveInvitation(invitation, {
-    reuseConnection: true,
-  })
+  let connectionRecord: ConnectionRecord | undefined, outOfBandRecord: OutOfBandRecord
```
TimoGlastra, line 264:
```suggestion
  let connectionRecord: ConnectionRecord | undefined
  let outOfBandRecord: OutOfBandRecord
```

packages/agent/src/parsers.ts
```diff
@@ -259,12 +261,21 @@ export async function receiveOutOfBandInvitation(
...
+
+  try {
+    ;({ connectionRecord, outOfBandRecord } = await agent.oob.receiveInvitation(invitation, {
+      reuseConnection: true,
+    }))
```
TimoGlastra, line 269: This is not really cleaner, but I really don't like the weird syntax when you destructure and reasign let values

```suggestion
  try {
    const receiveInvitationResult = await agent.oob.receiveInvitation(invitation, {
      reuseConnection: true,
    })
    connectionRecord = receiveInvitationResult.connectionRecord
    outOfBandRecord = receiveInvitationResult.outOfBandRecord
```

TimoGlastra (approved):

PR #70, opened 2023-09-29
chore: update version to 1.1.1

TimoGlastra:


berendsliedrecht (approved):
Thats smart! Will do the same for ID Crypt.

PR #71, opened 2023-10-02
feat: support mattr launchpad proof requests

TimoGlastra:
NOTE: very hacky

Uses packed versions of
- https://github.com/Sphereon-Opensource/PEX/pull/128
- https://github.com/Sphereon-Opensource/SIOP-OID4VP/pull/61

Only supports the single credential request yet, as the multi-request from launchpad contains an invalid presentation definition.


TimoGlastra:
I will try to get this updated with all the changes to credo. Would be nice to have credential selection in the wallet

berendsliedrecht:
@TimoGlastra Is this something we would like to add in the wallet right now? With some of the work we will do now, it might get even more out of sync.

TimoGlastra:
I have a local branch that has it 95% complete. I'll see to get it updated. I think the main thing was that the UX didn't work very nicely

PR #76, opened 2023-10-27
feat: Added support for url query inside the didcomm deeplink

Tommylans:
Not tested yet.


packages/agent/src/parsers.ts
```diff
@@ -346,12 +347,15 @@ export async function tryParseDidCommInvitation(
   invitationUrl: string
 ): Promise<OutOfBandInvitation | null> {
   try {
+    const parsedUrl = queryString.parseUrl(invitationUrl)
+    const updatedInvitationUrl = (parsedUrl['url'] as string | undefined) ?? invitationUrl
```
TimoGlastra, line 351: I think we should maybe call this oobUrl=?

BTW. For the deeplink in Paradym, we can just use https://paradym.id right, as we link to our own wallet?

So if the button says Open in Paradym Wallet, we don't really need the didcomm:// prefix?

Ofc. Good for other wallets

Tommylans: I'll update it, so it will work with the `didcomm://?oobUrl=` deeplink the https deeplinking needs a bit more work.

TimoGlastra (approved):

PR #82, opened 2023-11-22
feat: Wallet part of the https deeplinking

Tommylans:
I couldn't tested yet on Android for some reason I can't get it to start.

### DIFF
--- a/apps/expo/app/_layout.tsx
+++ b/apps/expo/app/_layout.tsx
@@ -143,6 +143,11 @@ export default function HomeLayout() {
           <NoInternetToastProvider>
             <DeeplinkHandler>
               <Stack screenOptions={{ headerShown: false }}>
+                {/*Workaround: To prevent the routing of the deeplinks*/}
+                <Stack.Screen name="invitation/[id]" redirect />
+                <Stack.Screen name="https/[...dummy]" redirect />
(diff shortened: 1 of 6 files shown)


apps/expo/app/_layout.tsx
```diff
@@ -63,9 +63,9 @@ export default function HomeLayout() {
       })
       if (!walletKey) return
 
-      const agent = await initializeAgent(walletKey).catch(() => {
-        toast.show('Could not initialize agent.')
+      const agent = await initializeAgent(walletKey).catch((e) => {
```
janrxyz, line 66:
```suggestion
      const agent = await initializeAgent(walletKey).catch(() => {
```

packages/agent/src/parsers.ts
```diff
@@ -42,6 +42,7 @@ export enum QrTypes {
   OPENID = 'openid://',
   OPENID_VC = 'openid-vc://',
   DIDCOMM = 'didcomm://',
+  HTTPS = 'https://',
```
janrxyz, line 45: Is this used anywhere?

janrxyz (commented):
Maybe clarify a bit more in a comment in _layout.tsx what exactly this does (not a lot just some more info besides that this is a workaround for deeplinking).

janrxyz (approved):

PR #97, opened 2024-04-07
feat: support openid invitations and deeplinking improvements

TimoGlastra:
This PR adds support for deeplinks and invitation urls for OpenID4VC created by Paradym.

It also expanded the parsing logic to be as flexible as possible. That means that:
- if `credential_offer` or `credential_offer_uri` query params, or any of the open id issuance schemes are used in a deeplink or a QR data url we scan we assume it's an openid credential offer
- if `request` or `request_uri` query params, or an of the open id  verification schemes are used in a deeplink or a QR data url we scan we assume it's an openid presentation request
- If `oobUrl`, `oob`, `c_i`, `d_m` query params or the `didcomm` schema is used in a deeplink or a qr data url we scan we assume it's didcomm invitation

If it can't be detected from the url alone we fetch the URL and then do anoter set of checks to see what type of invitation it is.

I also made the DIDcomm flow more similar to the openid flow, not showing the QR scanning too long, but redirecting to a loading screen. This also makes the deeplink flow better as it now doesn't do a lot of processing in the background which you can't see

I also added a general umatched route and just render `null` here. The deeplink handler resets the routing state now on every deeplink so we always route to the correct state (with the downside that you lose your navigation state when you deeplink into an active app, but with the three screen we have that's really a non-issue at the moment IMO). This means that ANY deeplink url (with random paths) now works, and if it isn't a deeplink we support it will just only reset your navigation state (so you'll be on the home screen)..

See attached videos for example flows.

https://github.com/animo/paradym-wallet/assets/23165168/30adc7a9-a664-4dc2-9100-e7d998d72926

https://github.com/animo/paradym-wallet/assets/23165168/3457d49e-e89a-4b6d-a9c7-e7eb19e69092

https://github.com/animo/paradym-wallet/assets/23165168/fa76198c-f84f-4c3b-9c98-bc5af06f61dd


apps/expo/app/[...unmatched].tsx
```diff
@@ -0,0 +1,16 @@
+import { usePathname, useGlobalSearchParams } from 'expo-router'
+
+// NOTE: for all unmatched routes we render null, as it's good chance that
+// we got here due to deep-linking, and we already handle that somewhere else
+export default () => {
```
Comment (Tommylans) [inline apps/expo/app/[...unmatched].tsx:5]: Quite a nicer workaround 👍

packages/agent/src/invitation/parsers.ts
```diff
@@ -0,0 +1,164 @@
...
+      // Parse invitation supports legacy connection invitations, oob invitations, and
+      // legacy connectionless invitations, and will all transform them into an OOB invitation.
+      const parsedInvitation = await agent.oob.parseInvitation(updatedInvitationUrl)
+
+      agent.config.logger.debug(`Parsed didcomm invitation with id ${parsedInvitation.id}`)
+      return {
+        success: true,
+        result: parsedInvitation,
+      } as const
+    } else {
```
Tommylans, line 106: Else isn't really needed here

Tommylans (approved):
Really nice solution for the deeplink also the rest is really clear what it does 🚀

PR #98, opened 2024-04-11
fix: key generation

Tommylans:


TimoGlastra (approved):

Tommylans:
Need to test it a bit more but after that merge and release

Tommylans:
Tested multiple cases and it does what it's meant to do.

PR #99, opened 2024-04-11
feat: Decline problem report

Tommylans:
Implementation for the decline problem report in the Paradym wallet.

### DIFF
--- a/packages/agent/src/hooks/useAcceptDidCommPresentation.ts
+++ b/packages/agent/src/hooks/useAcceptDidCommPresentation.ts
@@ -183,7 +183,7 @@ export function useAcceptDidCommPresentation(proofExchangeId: string) {
   return {
     acceptPresentation: acceptMutateAsync,
     declinePresentation: declineMutateAsync,
-    status: acceptStatus,
+    acceptStatus,
     proofExchange,
     submission: data,
     verifierName: connection?.theirLabel,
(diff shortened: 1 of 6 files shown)


packages/agent/src/hooks/useAcceptDidCommPresentation.ts
```diff
@@ -157,9 +157,33 @@ export function useAcceptDidCommPresentation(proofExchangeId: string) {
...
+        .pipe(
+          // Correct record with id and state
+          filter(
+            (event) =>
+              event.payload.proofRecord.id === proofExchangeId &&
+              [ProofState.Declined].includes(event.payload.proofRecord.state)
+          ),
+          // 10 seconds to complete exchange
+          timeout(10000),
+          first()
```
TimoGlastra, line 174: I think the method call will wait with returning until this state is reached, so i think we can remove it

TimoGlastra (approved):
Nice! How about credential decline?

packages/agent/src/hooks/useAcceptDidCommPresentation.ts
Tommylans: But it's quite nice to have this functionality right just like the accept action. And the state is updated locally in afj right for the record or does it have to wait for the external party?

Tommylans:
> Nice! How about credential decline?

Yes, added the credential decline now 😄

PR #102, opened 2024-04-15
fix: dislosed fields

TimoGlastra:
Properly handles showing the disclosed fields. For now it shows all fields that will be disclosed (like issuer, issuedAt, ec..). When we do a new iteration of the design we can show it as somethign like "In addition the following metadata is shared: Issuer, ... )"
IMG_6279


Tommylans (approved):
Looks good

PR #103, opened 2024-04-16
fix: sd-jwt/decode resolution

auer-martin:


TimoGlastra:
Can you also upgrade the version so we can release it ?

TimoGlastra (approved):

PR #104, opened 2024-04-16
fix: background color check

janrxyz:


packages/app/components/CredentialCard.tsx
```diff
@@ -50,11 +50,22 @@ export default function CredentialCard({
...
+    if (backgroundImage?.url) return '$transparent'
+    return bgColor ?? '$grey-900'
+  }
+
   return (
     <XStack
       shadow={shadow}
       br="$8"
-      bg={!backgroundImage ? bgColor ?? '$grey-900' : '$transparent'}
+      bg={getBgColor()}
```
TimoGlastra, line 68: No need really in it being a function right? Could just be a const assignment?

TimoGlastra (approved):

TimoGlastra:
Can you increase the version also so we can immediately release and publish?

PR #113, opened 2024-06-25
feat: credential selection in proof request

TimoGlastra:
This supersedes #71 and adds credential selection. It's a bit meh as you can't really see the credential you're selection. 

So I had two potential ideas:
- merge as is (with the known UX flaws)
- disable credential selection still in the UI for now, but then we at least have the logic in place to add it in later

Works with both DIDComm and OpenID4VC.

@berendsliedrecht @janrtvld WDYT?

https://github.com/animo/paradym-wallet/assets/23165168/298474bb-f76f-4e03-9642-d2aeed3cf931


TimoGlastra:
As you can see in the last commit, I disabled it for now in the UI. We can enable it again soon when we reworked the UX / design for selecting a credential

packages/agent/src/hooks/useDidCommPresentationActions.ts
```diff
@@ -46,90 +48,122 @@ export function useDidCommPresentationActions(proofExchangeId: string) {
...
-              issuerName: credentialDisplayMetadata?.issuerName ?? 'Unknown',
-              requestedAttributes: attributeNames,
-            })
-          }
+          const attributeArray = attributes[groupName] as AnonCredsRequestedAttributeMatch[]
+
+          submission.entries.push({
+            inputDescriptorId: groupName,
+            isSatisfied: attributeArray.length >= 1,
+            name: groupName, // TODO
```
janrtvld, line 73: What is this TODO?

packages/agent/src/hooks/useDidCommPresentationActions.ts
```diff
@@ -46,90 +48,122 @@ export function useDidCommPresentationActions(proofExchangeId: string) {
...
+            credentials: attributeArray.map((attribute) => {
+              const credentialExchange = credentialExchanges.find((c) =>
+                c.credentials.find((cc) => cc.credentialRecordId === attribute.credentialId)
+              )
+              const credentialDisplayMetadata = credentialExchange
+                ? getDidCommCredentialExchangeDisplayMetadata(credentialExchange)
+                : undefined
+
+              return {
+                name: groupName, // TODO: humanize string? Or should we let this out?
```
janrtvld, line 83: I often humanize in the UI component to keep it separated.

packages/app/src/features/notifications/components/PresentationNotificationScreen.tsx
```diff
@@ -21,79 +36,145 @@ export function PresentationNotificationScreen({
...
+          <Stack bg="$white" pb="$4">
+            {currentSubmissionEntry?.credentials.map((c, credentialIndex) => (
+              <CredentialRowCard
+                onPress={() => {
+                  onSelectCredentialForInputDescriptor(currentSubmissionEntry.inputDescriptorId, credentialIndex)
+                  setChangeSubmissionCredentialIndex(-1)
+                }}
+                // The index is stable enough here
+                // biome-ignore lint/suspicious/noArrayIndexKey: <explanation>
+                key={credentialIndex}
```
janrtvld, line 168: Does the credential not have an id we can use? I learned my lesson a few weeks back when debugging an index used as a key for way too long.

packages/ui/src/panels/Sheet.tsx
janrtvld, line 1: What was wrong with the tamagui sheet?

janrtvld (approved):

packages/agent/src/hooks/useDidCommPresentationActions.ts
TimoGlastra: The name is the group name if no credentials are found (as that' the only value we have), which can be quite ugly. But this was like this before as well

packages/app/src/features/notifications/components/PresentationNotificationScreen.tsx
TimoGlastra: I think we can add it. But we don't have it right now in this model. But yeah no reason to not use the id

packages/ui/src/panels/Sheet.tsx
TimoGlastra: It was a year ago that I implemented this so ... TBH I don't know anymore. I can vaguely remember that It was just buggy and I've had good experience with this sheet (as it's rendered as native sheet on iOS)

TimoGlastra: I think it had to do also with scroll. If you have a lot of credentials, it will become a scroll the sheet, which also didn't work very well with the tamagui sheet (if i remember correctly)

berendsliedrecht:
@TimoGlastra do you need an additional review on this?

TimoGlastra:
No it's ok 👍

TimoGlastra:
I've updated this with the grouping, but have improved upon it a bit. We prefer grouping by __CREDENTIAL__, as it's what we use in Paradym and otherwise if all the matches for a requested entry match. In most cases I think this should suffice. And we will filter the matches so it correctly shows the options if selection is enabled that can be used for all credentials in the group.

There is one case that is not catched by this, and this is when you request a predicate and values from the same credential basically, but without the __CREDENTIAL__ prefix in group name. This can gives multiple separate optoins ,but I feel this is more a limitation of AnonCreds proof requests, which we have now sovled using the __CREDENTIAL__ prefix

There has been some pushback on adding this feature to AnonCreds v1, so I think our 'workaround' is fine here. https://github.com/hyperledger/anoncreds-rs/issues/292

TimoGlastra:
Adding a "received on XXX" will probably already help a lot with being able to distinguish credentials when selecting

PR #118, opened 2024-07-04
feat: Merge anoncreds proof request credentials

Tommylans:
Made it so it favors the way of combining the credentials in Paradym, after that it will combine them based on the credentialId. And if that isn't available it will still use the groupname. 

So the groupname will be used when the credential is not available in the wallet and the proof request didn't came from Paradym.

image

### DIFF
--- a/packages/agent/src/hooks/useDidCommPresentationActions.ts
+++ b/packages/agent/src/hooks/useDidCommPresentationActions.ts
@@ -68,8 +68,10 @@ export function useDidCommPresentationActions(proofExchangeId: string) {
 
           const firstMatch = attributeArray[0]
 
-          // When the credentialId isn't available, we use the groupName as the key but it will result in multiple entries in the view. But I think it's not an easy task to merge them
-          const credentialKey = firstMatch?.credentialId ?? groupName
+          // When the credentialId isn't available and there is no __CREDENTIAL__ in the groupName, we use the groupName as the key but it will result in multiple entries in the view. But I think it's not an easy task to merge them
+          const credentialKey =
+            firstMatch?.credentialId ??
+            (groupName.includes('__CREDENTIAL__') ? groupName.split('__CREDENTIAL__')[0] : groupName)
 
           if (!firstMatch) {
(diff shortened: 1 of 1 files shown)


TimoGlastra (approved):
How does this work with credential selection? If there's multiple
credential available, changing one should change all

So we migut want to do something smarter, also looking at requirements. So it works generally

Tommylans:
> How does this work with credential selection? If there's multiple credential available, changing one should change all
> 
> So we migut want to do something smarter, also looking at requirements. So it works generally

But what would make sense? When you click on it that you get the selector of #113 and that they explode again so they get splitted again? 

Or should we always look for the `credentialId` and ignore the `__CREDENTIAL__` because that implementation doesn't look at the selected credential.

packages/agent/src/hooks/useDidCommPresentationActions.ts
```diff
@@ -94,8 +111,11 @@ export function useDidCommPresentationActions(proofExchangeId: string) {
           // This should probably be fixed in AFJ.
           const firstMatch = predicateArray[0]
 
+          // When the credentialId isn't available, we use the groupName as the key but it will result in multiple entries in the view. But I think it's not an easy task to merge them
+          const credentialKey = firstMatch?.credentialId ?? groupName
```
TimoGlastra, line 115: Maybe if no credentialId is available we match based on the CREDENTIAL_ prefix like you had before as fallback?

Otherwise there's not an easy way to group them

TimoGlastra:
I was a bit confused, but see now that https://github.com/animo/paradym-wallet/issues/113 hasn't been merged yet 🤦 in my head it was.

But your approach probably will also work with that pr, so i think we're good 👍

packages/agent/src/hooks/useDidCommPresentationActions.ts
Tommylans: Sure brought it back 😄

TimoGlastra (approved):

PR #119, opened 2024-07-13
feat: react native keychain setup

TimoGlastra:
Sets up react native keychain, keeps expo-secure-store for now until we have time to also update the paradym wallet

Haven't tested it yet, as I first need to set up everything for PIN unlock and biometrics, but the code to store SALT / wallet key, derive wallet key, etc.. is there. 

Will now work on the onboarding flow so we can correctly set up the wallet key etc..


packages/secure-store/legacyUnlock.ts
```diff
@@ -0,0 +1,41 @@
+import { ariesAskar } from '@hyperledger/aries-askar-shared'
+import * as SecureStore from 'expo-secure-store'
+
+const STORE_KEY_LEGACY = 'wallet-key' as const
+const STORE_KEY_RAW = 'paradym-wallet-key-raw' as const
```
berendsliedrecht, line 5: paradym specific identifier

packages/secure-store/secure-unlock/saltStore.ts
```diff
@@ -0,0 +1,35 @@
+import * as Keychain from 'react-native-keychain'
+import { type KeychainOptions, getKeychainItemById, storeKeychainItem } from '../keychain'
+
+const saltStoreBaseOptions: KeychainOptions = {
+  /* Salt can be accessed on this device */
+  accessible: Keychain.ACCESSIBLE.WHEN_UNLOCKED_THIS_DEVICE_ONLY,
+}
+
+const SALT_ID = (version: number) => `PARADYM_WALLET_SALT_${version}`
```
berendsliedrecht, line 9: Paradym specific again

packages/secure-store/secure-unlock/walletKeyDerivation.ts
```diff
@@ -0,0 +1,29 @@
...
+}
+
+/**
+ * Generate 32 byte key crypto getRandomValues.
+ *
+ * @see https://github.com/LinusU/react-native-get-random-values
+ * @see https://developer.mozilla.org/en-US/docs/Web/API/Crypto/getRandomValues
+ */
+export function generateSalt(): string {
+  return crypto.getRandomValues(new Uint8Array(32)).join('')
```
berendsliedrecht, line 28: Is ths from aries askar of react-native-random-values?

packages/secure-store/secure-unlock/walletKeyStore.ts
```diff
@@ -0,0 +1,98 @@
...
+  securityLevel: Keychain.SECURITY_LEVEL.SECURE_HARDWARE,
+
+  /* Android Only. Ensure wallet key is protected using RSA storage type, this is the only storage type supporting biometrics. */
+  storage: Keychain.STORAGE_TYPE.RSA,
+
+  /* Ensure wallet key is protected by biometrics. It is not possible to fallback to the device passcode if the biometric authentication failed. */
+  authenticationType: Keychain.AUTHENTICATION_TYPE.BIOMETRICS,
+}
+
+const WALLET_KEY_ID = (version: number) => `PARADYM_WALLET_KEY_${version}`
```
berendsliedrecht, line 30: Paradym specific identifier again.

packages/secure-store/secure-unlock/walletKeyStore.ts
```diff
@@ -0,0 +1,98 @@
...
+
+  /**
+   * We only support biometrics secured storage of the wallet key
+   */
+  const supportedBiometryType = await Keychain.getSupportedBiometryType(walletKeyStoreBaseOptions)
+  if (!supportedBiometryType) {
+    return false
+  }
+
+  return true
```
berendsliedrecht, line 73:
```suggestion
  return Keychain.getSupportedBiometryType(walletKeyStoreBaseOptions)
```

packages/secure-store/secureUnlock.ts
```diff
@@ -0,0 +1,47 @@
...
+import {
+  canUseBiometryBackedWalletKey,
+  getWalletKeyUsingBiometrics,
+  storeWalletKey,
+} from './secure-unlock/walletKeyStore'
+
+// TODO: how to version?
+const version = 1
+
+export { canUseBiometryBackedWalletKey }
```
berendsliedrecht, line 13: The functions, mainly their names, confuse me a bit and to me it is not really clear when to use which one. Also, since they are all exported, the API will get quite confusing I imagine.

Would it be possible to make it into an object/class with static methods, so we use it as a namespace?


i.e.

```js
// Wallet as a placeholder
Wallet.getSalt()
Wallet.createSaltAndIncrementVersion()
Wallet.deriveWalletKeyFromPin()
Wallet.unlockWallet() // do we need to mention using biometrics here?
```

berendsliedrecht (approved):
Nice work! COuple of small points.

apps/funke/app.config.js
```diff
@@ -52,13 +52,14 @@ const config = {
   updates: {
     fallbackToCacheTimeout: 0,
   },
-  plugins: ['expo-font', 'expo-secure-store', 'expo-router'],
+  plugins: ['expo-font', 'expo-router'],
```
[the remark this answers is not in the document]
TimoGlastra: For the funke app it doesn't really matter yet. Also -- the plugin is for configuration of the Face id description, it doesn't impact the bundling.

packages/secure-store/legacyUnlock.ts
TimoGlastra: Yes, but this need to stay the same as this it the legacy unlock (currnetly used by paradym wallet). We can't just change the key

packages/secure-store/secure-unlock/saltStore.ts
TimoGlastra: I feel it is ok to use PARADYM_WALLET as a prefix here. As the 'core' I'd see as paradym wallet. I didn't want to go fully generic "WALLET_SALT"

packages/secure-store/secure-unlock/walletKeyDerivation.ts
TimoGlastra: React Native Random Values

packages/secure-store/secure-unlock/walletKeyStore.ts
TimoGlastra: see my previous comment. Any suggestions on what you'd use?

packages/secure-store/secure-unlock/walletKeyStore.ts
TimoGlastra: lol :)

packages/secure-store/secureUnlock.ts
TimoGlastra: Yes we can do that, I'll do it when I integrate though, is that OK? Then I can play a bit more with how the API is easy to use. 

Re mentioning biometrics, I did want to mention it as there's a difference between unlocking the wallet using biometrics (no PIN/SALT needed) or using a PIN. As the secure store only deals with deriving and storing of wallet key, it doesn't actually open the wallet.

So it should be more "retrieveWalletKeyUsingBiometrics" and "deriveWalletKeyUsingPin", what do you think of these names?

packages/secure-store/secure-unlock/walletKeyStore.ts
berendsliedrecht: We'd likely need a config json for each app which defines the theme stuff but also items like this. Might be a bit overkill though.

packages/secure-store/secureUnlock.ts
berendsliedrecht: Names are good!

packages/secure-store/secure-unlock/walletKeyStore.ts
TimoGlastra: Yeah I don't really see the added value of that. Then I'd prefer going with `WALLET_KEY_${version}`

berendsliedrecht: Okay, paradym prefix is also fine for me. It does not really matter in the end.

PR #120, opened 2024-07-14
feat: add storybook

TimoGlastra:
Added storybook so it's a bit easier to develop components in the app. As tamgui also supports Web, we can run storybook in Web as long as we don't use any code that doesn't support react-native-web (most of expo supports it).

We do need to move more UI components to the `@pacakge/ui` and have the `@package/app` mostly focused on the logic. This way we can develop UI without having to run the app. See example in video below

https://github.com/user-attachments/assets/c970d3b6-7cce-4b78-91ad-0e9def5e9422


TimoGlastra:
it also runs on-device: 

https://github.com/user-attachments/assets/5ee8b6c2-0b90-4edd-972d-1697332a8319

janrtvld (approved):

PR #121, opened 2024-07-14
feat: custom config and font for funke

TimoGlastra:
Builds on top of #120. Adds custom config for the Funke wallet with custom primary color and fonts. The font setup has been improved in general by loading them in the app config, which pre-loads them and means we don't have to load the fonts on startup.

Paradym wallet still uses Inter, Funke wallet uses Raleway.

@janrtvld Maybe you can help a bit with configuring the theme and colors / sizes for the Funke app. I only configured the primary color now.


TimoGlastra:
This branch also contains some fixes for #120 . so we either need to merge at the same time, or merge this PR into #120 first

apps/funke/app.config.js
```diff
@@ -52,7 +52,21 @@ const config = {
...
-  plugins: ['expo-font', 'expo-secure-store', 'expo-router'],
+  plugins: [
+    [
+      'expo-font',
+      {
+        fonts: [
+          '../../node_modules/@expo-google-fonts/raleway/Raleway_400Regular.ttf',
+          '../../node_modules/@expo-google-fonts/raleway/Raleway_500Medium.ttf',
+          '../../node_modules/@expo-google-fonts/raleway/Raleway_600SemiBold.ttf',
+          '../../node_modules/@expo-google-fonts/raleway/Raleway_700Bold.ttf',
```
janrtvld, line 63: We actually use 'Raleaway' for only the titles and headings. The body font is 'Open sans'.

apps/funke/tamagui.config.ts
```diff
@@ -1,13 +1,46 @@
...
+      ...tokens.color,
+    },
+  },
+})
+
+type ConfIg = typeof config
 declare module 'tamagui' {
   // eslint-disable-next-line @typescript-eslint/no-empty-interface
-  interface TamaguiCustomConfig extends Conf {}
+  interface TamaguiCustomConfig extends ConfIg {}
```
janrtvld, line 43: Why this change?

apps/funke/app.config.js
TimoGlastra: Okay will update 👍

apps/funke/tamagui.config.ts
TimoGlastra: will revert

TimoGlastra: Reverted in another PR

apps/funke/app.config.js
TimoGlastra: I will do this in a separate PR, easier and it's not an issue for the Funke wallet

PR #123, opened 2024-07-15
feat: tamagui funke config

janrtvld:


apps/funke/app.config.js
```diff
@@ -6,14 +6,17 @@ const variants = {
   development: {
     bundle: '.dev',
     name: ' (Dev)',
+    mediatorDid: 'did:web:mediator.dev.paradym.id',
```
TimoGlastra, line 9: Funke has no didcomm setup, so this is not needed I think. did you get an error?

TimoGlastra (approved):

packages/app/src/provider/Provider.tsx
```diff
@@ -1,17 +1,28 @@
...
 import { useColorScheme } from 'react-native'
 
+import type { PropsWithChildren } from 'react'
 import { ToastViewport } from './ToastViewport'
 
 const queryClient = new QueryClient()
 
-export function Provider({ children, ...rest }: TamaguiProviderProps) {
+interface ProviderProps extends TamaguiProviderProps {
+  tamaguiConfig: TamaguiInternalConfig
```
TimoGlastra, line 13: Tamagui already accepts a prop `config`. should we omit that?

TimoGlastra (approved):

PR #125, opened 2024-07-16
feat: use hardware backed keys

berendsliedrecht:
- Just waiting on a new release of credo.


packages/agent/src/invitation/handler.ts
```diff
@@ -96,11 +98,20 @@ export const receiveCredentialFromOpenId4VciOffer = async ({
           didMethod = 'key'
         }
 
+        const key = await agent.wallet
+          .createKey({
+            keyType,
+            keyBackend: KeyBackend.SecureElement,
+          })
+          .catch(() => {
+            throw new Error(`Could not create a hardware-backed key for keytype: '${keyType}'. Only P-256 is supported`)
```
berendsliedrecht, line 107: How do we want to deal with this? Do we want to make sure secure element is only used for specific credentials? Only using secure env keys makes the wallet useless for phones without one.

TimoGlastra:
We can add patches for now (i had to do this for a separate branch recently). Pushed it to this PR

packages/agent/src/invitation/handler.ts
TimoGlastra: I think we should have a way to detect PID/QEAA usage, and if so, require hardware. But otherwise not use it. 

Not sure how yet ... Maybe it can be an app.json config for now (like mediatorDid). As for Funke we now only have to deal with PIDs

berendsliedrecht: How do we want to do that? Maybe the VCT scheme in the app.json?

TimoGlastra: So I think for now just make it a true/false whether to use hardware keys. If yes, you can only use hardware P-256. If no, you can only use software keys

TimoGlastra (approved):

PR #126, opened 2024-07-23
feat: onboarding screens in storybook

TimoGlastra:
This adds the screens for onboarding flow. As a next steps I want to integrate them into the funke wallet, and also move them out of the stories.

So it's a bit of an in-between state, but I want to merge it in iterations.

See the video for the implemented flow

https://github.com/user-attachments/assets/7bc17703-197d-462e-b558-db569d665602


apps/funke/tamagui.config.ts
```diff
@@ -45,9 +46,21 @@ const config = createTamagui({
   fonts: {
     default: fontOpenSans,
     heading: fontRaleway,
+    // Somehow adding body font gives build errors?!
+    // body: fontOpenSans,
```
berendsliedrecht, line 50: 😭

apps/storybook/components/funke/IdCardPinScreen.stories.tsx
```diff
@@ -0,0 +1,243 @@
...
+
+  return (
+    <Page gap="$6" justifyContent="space-between">
+      <ProgressBar value={25} />
+      <YStack flexBasis={0} flexGrow={2} flexShrink={1} justifyContent="space-between">
+        <Heading variant="title">{title}</Heading>
+        <XStack display={state === 'enterPin' ? 'flex' : 'none'} gap="$3" justifyContent="center">
+          {pin.map((digit, index) => (
+            // biome-ignore lint/suspicious/noArrayIndexKey: index is the correct key here
+            <YStack key={index} maxWidth={35} flex-1 justifyContent="center">
```
berendsliedrecht, line 144: IIRC @janrtvld this caused some very very weird errors in the idcrypt pin screen right?

apps/storybook/components/funke/PinScreen.stories.tsx
```diff
@@ -0,0 +1,94 @@
...
+  return (
+    <Page gap="$6" justifyContent="space-around">
+      <ProgressBar value={25} />
+      <YStack gap="$2" flex={3}>
+        <Heading variant="title">{title}</Heading>
+        {subtitle && <Paragraph>{subtitle}</Paragraph>}
+      </YStack>
+      <YStack flex={9} mt="$5" onPress={focusInput}>
+        <XStack justifyContent="center" gap="$2">
+          {new Array(pinLength).fill(0).map((_, i) => (
```
berendsliedrecht, line 39: Same here.

berendsliedrecht (approved):

apps/storybook/components/funke/BiometricsModal.stories.tsx
```diff
@@ -0,0 +1,87 @@
...
+      disableDrag
+      dismissOnOverlayPress={false}
+    >
+      <Sheet.Overlay
+        animation="lazy"
+        enterStyle={{ opacity: 0 }}
+        exitStyle={{ opacity: 0 }}
+        backgroundColor="$black"
+        opacity={0.2}
+      />
```
janrtvld, line 43: Nit but this is my preset for the tamagui sheet. I find the default a bit too dark and slow.

```
<Sheet.Overlay
          bc="#00000033"
          animation="quick"
          enterStyle={{ opacity: 0 }}
          exitStyle={{ opacity: 0 }}
        />
```

apps/storybook/components/funke/BiometricsModal.stories.tsx
```diff
@@ -0,0 +1,87 @@
...
+            <Heading variant="h2">Biometrics is required</Heading>
+            <Paragraph variant="text">
+              Securing with your biometrics is required for receiving your digital identity
+            </Paragraph>
+          </YStack>
+          <Button.Solid onPress={onAcceptBiometrics} mx="$4" mb="$4">
+            Activate Biometrics
+          </Button.Solid>
+        </YStack>
+      </Sheet.Frame>
```
janrtvld, line 59: So ideally we have the sheet itself take the height of the content. And then we use padding around the Icon to scale the light-shaded section. 

Then use 'gap' to separate the sections from eachother. 

The text section can then also take the height based on the content. Using fixed heights will mess it up if your text is longer on smaller devices (or because of text changes later.

apps/storybook/components/funke/FirstLaunchScreen.stories.tsx
```diff
@@ -12,42 +11,26 @@ const FirstLaunchScreen = () => {
...
-          {/* todo: grey-5 */}
-          <Heading variant="h1" color="#282C37">
-            Ausweis Wallet
-          </Heading>
-          {/* todo: 1 is a bit too small, maybe just hardcode */}
-          <Separator borderWidth="$1" borderRadius="$1" borderColor="$primary-500" width="$3" />
-          {/* todo: grey-4 */}
-          <Heading variant="h1" color="#656974">
+          <Heading variant="title">Ausweis Wallet</Heading>
+          <Separator borderWidth={3} borderRadius={3} borderColor="$primary-500" width="$4" />
```
janrtvld, line 23: Can you use token values for the borderWidth and radius as well? Or did they not work.

"$1" or "$1.5" maybe.

```suggestion
          <Separator borderWidth="$1" borderRadius="$1" borderColor="$primary-500" width="$4" />
```

apps/storybook/components/funke/IdCardPinScreen.stories.tsx
```diff
@@ -0,0 +1,243 @@
...
+  complete: <HeroIcons.ShieldCheck color="$white" />,
+} as const
+
+interface IdCardProps {
+  icon: keyof typeof iconMapping
+  issuerImage: string
+  userName?: string
+}
+
+function IdCard({ icon, issuerImage, userName }: IdCardProps) {
```
janrtvld, line 50: We're going to use this in more places so i'd extract this into it's own component.

apps/storybook/components/funke/IdCardPinScreen.stories.tsx
```diff
@@ -0,0 +1,243 @@
...
+          </Paragraph>
+        </YStack>
+        <Stack>
+          <Image src={issuerImage} width={75} height={75} resizeMode="contain" />
+        </Stack>
+      </XStack>
+      <XStack justifyContent="flex-end">
+        {/* TODO: background color variables */}
+        <Circle size="$3" backgroundColor={icon === 'complete' ? '#C9D6BD' : '#02010033'}>
+          {iconMapping[icon]}
```
janrtvld, line 80: Might be okay for now to just have a grey circle with a slight opacity so that it blends a bit with where we put it. (no need for colours)

apps/storybook/components/funke/IdCardPinScreen.stories.tsx
```diff
@@ -0,0 +1,243 @@
...
+    if (firstEmptyIndex !== -1 && firstEmptyIndex < index) {
+      inputRefs.current[firstEmptyIndex]?.focus()
+    } else {
+      inputRefs.current[index]?.focus()
+    }
+  }
+
+  return (
+    <Page gap="$6" justifyContent="space-between">
+      <ProgressBar value={25} />
```
janrtvld, line 138: I've split it into 3 sections with the most recent draft.

```suggestion
      <ProgressBar value={33} />
```

apps/storybook/components/funke/IdCardPinScreen.stories.tsx
```diff
@@ -0,0 +1,243 @@
...
+    if (index < pin.length - 1) {
+      programmaticFocus.current = true
+      inputRefs.current[index + 1]?.focus()
+    } else if (index === pin.length - 1) {
+      inputRefs.current[index]?.blur()
+      onPinComplete(pin.join(''))
+    }
+  }
+
+  const focusInput = (index: number) => {
```
janrtvld, line 118: Could we not make it so this can never be unfocussed?

apps/storybook/components/funke/IdCardPinScreen.stories.tsx
```diff
@@ -0,0 +1,243 @@
...
+                accessible={true}
+                aria-label={`Pin digit ${index + 1}`}
+                accessibilityHint="Enter the digit of the pin"
+              />
+              <Stack borderBottomWidth={1} borderBottomColor="$grey-900" width="100%" />
+            </YStack>
+          ))}
+        </XStack>
+      </YStack>
+      <Stack gap="$4" flexBasis={0} flexGrow={4} flexShrink={1}>
```
janrtvld, line 173: For the layout you use a lot of different flex methods which can become quite a mess if you want to structure a page layout.

My approach is always using justify-content to place the main elements of a stack (which you do nicely). And then use a combination of padding and gap to space the content elements. 

I would combine the progress bar and the title into 1 re-usable header for all the pages. And then have one YStack as the parent for the rest of your content. You can give this a flex grow of 1 to make it scale to all the space that is left on your screen.
Ideally you set the boundaries of your content on this stack, so probably a padding of '4' to make it stand out from the sides.

Then you have a stack with the pin input and the pin card. The spacing for these elements can be done with gap to make theme separated from each other. 

So then you only need flex grow and gap and sometimes some padding.

janrtvld: It might also be preference but I think it's much easier to read and debug then 3 different flex methods on every stack that is used. 

I think as long as you have the base layout (your main elements spread out using 'justify-content' and a base content stack with 'flex-grow') then you can still use it for your inner content.

apps/funke/tamagui.config.ts
```diff
@@ -45,9 +46,21 @@ const config = createTamagui({
   fonts: {
     default: fontOpenSans,
     heading: fontRaleway,
+    // Somehow adding body font gives build errors?!
```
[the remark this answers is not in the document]
TimoGlastra: No :/

apps/funke/tamagui.config.ts
```diff
@@ -45,9 +46,21 @@ const config = createTamagui({
...
+      ...tokens.color,
+
+      // Button
+      buttonHeight: 53,
+
+      // Button.Outline
+      buttonOutlineBackgroundColor: hexColors['grey-100'],
+      buttonOutlineBorderColor: '#ebedef',
+      buttonOutlineTextColor: hexColors['grey-900'],
+    },
```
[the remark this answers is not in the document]
TimoGlastra: > Also ideally we don't set fixed heights on buttons. I mostly used padding to resize it.

Ah okay, took your design too literally then :) 

> What button is this? Can't we do this in the button component itself?

The outline button. We can't really do it as it changes based on the config and it's very different for Paradym and Funke.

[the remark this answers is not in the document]
TimoGlastra: But if you have suggestions / concrete implementation proposal I'm all ears

apps/storybook/components/funke/BiometricsModal.stories.tsx
```diff
@@ -0,0 +1,87 @@
...
+}
+
+const BiometricsModal = ({ onAcceptBiometrics, open, onOpenChange }: BiometricsModalProps) => {
+  return (
+    <Sheet
+      open={open}
+      onOpenChange={onOpenChange}
+      // FIXME: hardcoded for good fit in storybook, but this probably won't work on mobile...
+      snapPoints={[520]}
+      snapPointsMode="constant"
```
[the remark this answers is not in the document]
TimoGlastra: Yeah this is a storybook thing, as the actual screen is higher than the displayed screen. I'll rework it when integrating on the actual page

apps/storybook/components/funke/BiometricsModal.stories.tsx
TimoGlastra: Yeah I need to look into this again on a proper RN screen. Sheet and Storybook was giving a lot of troubles.

apps/storybook/components/funke/FirstLaunchScreen.stories.tsx
TimoGlastra: Yes, but I think not everyone should be based on the size values right? as it will just look weird when the value of $1 changes

apps/storybook/components/funke/IdCardPinScreen.stories.tsx
TimoGlastra: It is it's own component?

apps/storybook/components/funke/IdCardPinScreen.stories.tsx
TimoGlastra: I'll think about it

apps/storybook/components/funke/IdCardPinScreen.stories.tsx
TimoGlastra: I spent an enormous amount of time on this, and there's a reason I did it exactly this way, as otherwise the content would jump around. 

I even opened an issue in tamagui for it: https://github.com/tamagui/tamagui/issues/2806

TimoGlastra: It seems it was an actual bug, and they already fixen my issue :)

apps/storybook/components/funke/IdCardPinScreen.stories.tsx
janrtvld: A separate ID Card UI component that can be re-used across multiple screens, as it's now only in this Pin screen.

apps/storybook/components/funke/FirstLaunchScreen.stories.tsx
janrtvld: Ah yes you're right. For borders I often used fixed values as they don't need to change ever.

apps/funke/tamagui.config.ts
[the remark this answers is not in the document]
janrtvld: Ah this is the button with the icon? And i'm guessing the icon makes it higher than the regular button with only text.

For one-time overrides like this i'd pass in the height as a prop where you use it. It's often better than cluttering your config with values (imo).

The colouring is harder. Again you could use a prop override, but you don't want to do that all the time. As you also have the height override, it might be better to just make a new button variant that fits the requirements.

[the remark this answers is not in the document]
TimoGlastra: > And i'm guessing the icon makes it higher than the regular button with only text.

No not necessarily . I just set the button height for all buttons as it was very specific in hte design and the same height everywhere. Didn't know it was meant to be set with padding. So will update it.


For the outline button with icon it is still important to set the height/width i guess to make it squre ( as most icons are not square). I'll look into a new button variant

apps/storybook/components/funke/IdCardPinScreen.stories.tsx
TimoGlastra: Ah yes, will do the extracting out of storybook soon

apps/storybook/components/funke/IdCardPinScreen.stories.tsx
TimoGlastra: I looked into it, and it's quite tricky with the all the input components, and also having to handle e.g. starting over with the pin.

apps/storybook/components/funke/IdCardPinScreen.stories.tsx
TimoGlastra: Okay cleaned it up quite a bit now. Based on your approach

PR #127, opened 2024-07-25
feat: Upgraded all packages for cheqd revocation support

Tommylans:
It's not well implemented yet. You can't really see that it is revoked it's just not selectable for the proof request. But it the package upgrades are needed to have the ability to fetch the revocation registry and the statuslist of cheqd.


TimoGlastra:
This PR can be closed, as all updates are already made with the Funke updates

PR #128, opened 2024-07-29
feat: pin_derived_eph_pub+priv

berendsliedrecht:


apps/funke/app/crypto/bPrime.ts
```diff
@@ -0,0 +1,34 @@
+import type { AgentContext } from '@credo-ts/core'
```
TimoGlastra, line 1: Everything within app is seen as routes, so I think we should add all this logic outside of `app` dir

packages/secure-store/kdf.ts
```diff
@@ -0,0 +1,41 @@
...
+ * Derive a hash from pin and salt. (which can be used to seed a key)
+ *
+ * Configuration based on recommended parameters defined in RFC 9106
+ * @see https://www.rfc-editor.org/rfc/rfc9106.html#name-parameter-choice
+ *
+ * returns a hex-encoded derived hash
+ *
+ */
+const derive = async (pin: string | Uint8Array, salt: string | Uint8Array): Promise<string> => {
+  const normalizedPin = typeof pin === 'string' ? pin : Buffer.from(pin).toString()
```
TimoGlastra, line 14: We normally use TypedArrayEncoder

packages/secure-store/secure-unlock/react-native-argon2.d.ts
TimoGlastra, line 1: I think this needs to stay

TimoGlastra (approved):

packages/secure-store/secure-unlock/react-native-argon2.d.ts
berendsliedrecht: Moved it to the root of the project, made more sense to me.

apps/funke/app/crypto/bPrime.ts
berendsliedrecht: I'll move it out, but now we just get a very messy root where `ios`, `node_modules`, `app` and now `crypto` is defined...

TimoGlastra: We can add a src directory

PR #129, opened 2024-07-30
store seed credential

berendsliedrecht:
Builds ontop of #125


packages/agent/src/storage/walletJsonStore.ts
```diff
@@ -0,0 +1,54 @@
...
+  const transformedValue = item.value === 'string' ? JSON.parse(item.value) : item.value
+
+  return transformedValue as T
+}
+
+export const walletJsonStore = {
+  store,
+  getById,
+  replace,
+}
```
TimoGlastra, line 54: We have agent.genericRecords for this. No need to use askar directory.

berendsliedrecht: Ah yeah makes sense, will swap.

packages/agent/src/invitation/handler.ts
```diff
@@ -35,44 +34,51 @@ import {
...
   if (!offerUri && data) {
     // FIXME: Credo only support credential offer string, but we already parsed it before. So we construct an offer here
     // but in the future we need to support the parsed offer in Credo directly
-    offerUri = `openid-credential-offer://credential_offer=${encodeURIComponent(JSON.stringify(data))}`
+    offerUri = `openid-credential-offer://credential_offer=${encodeURIComponent(
+      JSON.stringify(data)
+    )}`;
   } else if (!offerUri) {
-    throw new Error('either data or uri must be provided')
+    throw new Error("either data or uri must be provided");
```
TimoGlastra, line 67: why the formatting change?

berendsliedrecht: I still have prettier setup, will run biome again...just a mistake.

packages/agent/src/invitation/handler.ts
```diff
@@ -122,33 +118,6 @@ export const receiveCredentialFromOpenId4VciOffer = async ({
...
-          const didKey = DidKey.fromDid(didResult.didState.did)
-          verificationMethodId = `${didKey.did}#${didKey.key.fingerprint}`
-        }
-
-        return {
-          didUrl: verificationMethodId,
-          method: 'did',
-        }
-      }
-
```
TimoGlastra, line 151: why did you remove this? This is needed for the Paradym WAllet?

berendsliedrecht: I don't think there was a reason, was just a bit too focussed on Funke wallet. Will revert.

TimoGlastra (approved):

PR #133, opened 2024-08-06
chore: check vct scheme to choose hardware/software key

berendsliedrecht:

### DIFF
--- a/packages/agent/src/invitation/handler.ts
+++ b/packages/agent/src/invitation/handler.ts
@@ -142,11 +142,13 @@ export const receiveCredentialFromOpenId4VciOffer = async ({
   credentialConfigurationIdToRequest,
   accessToken,
   clientId,
+  pidSchemes,
 }: {
   agent: EitherAgent
   resolvedCredentialOffer: OpenId4VciResolvedCredentialOffer
   credentialConfigurationIdToRequest?: string
   clientId?: string
+  pidSchemes?: { sdJwtVcVcts: Array<string>; msoMdocNamespaces: Array<string> }
 
   // TODO: cNonce should maybe be provided separately (multiple calls can have different c_nonce values)
   accessToken: OpenId4VciRequestTokenResponse
@@ -182,6 +184,7 @@ export const receiveCredentialFromOpenId4VciOffer = async ({
       supportsAllDidMethods,
       supportsJwk,
       credentialFormat,
+      supportedCredentialId,
     }) => {
       // First, we try to pick a did method
       // Prefer did:jwk, otherwise use did:key, otherwise use undefined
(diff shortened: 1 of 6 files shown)


packages/agent/src/invitation/handler.ts
```diff
@@ -199,26 +201,15 @@ export const receiveCredentialFromOpenId4VciOffer = async ({
...
-              keyType,
-            })
-          })
-      } else {
-        key = await agent.wallet.createKey({
-          keyType,
-        })
-      }
+      // TODO: require support for mDoc namespace here as well
+      const vct = resolvedCredentialOffer.offeredCredentialConfigurations[offeredCredentialToRequest.id].vct
```
TimoGlastra, line 205: Use `supportedCredentialId` that is passed to this callback. This will help us when we add mdoc as well and request multiple creds.

Also -- vct is not always defined. So this check should be something like:

```
let shouldKeyBeHardwareBacked = false

const offeredCredentialConfiguration = offeredCredentialConfigurations[supportedCredentialId]

// TODO: use the enum for format, not sure what its' called)
if (offeredCredentialConfiguration.format === 'vc+sd-jwt' && pidSchemes?.sdJwtVcVcts.includes(offeredCredentialConfiguration.vct)) {
  shouldKeyBeHardwareBacked = true
}

```

Finally, I'm not sure why we need to call it pidSchemes. That's more implementation specific. Why not call it useHardwareKeySchemes? Or just pass `useHardwareKey: true/false` to this method? In receive pid use case we pass true, by default it's false

packages/agent/src/invitation/handler.ts
```diff
@@ -199,26 +202,26 @@ export const receiveCredentialFromOpenId4VciOffer = async ({
...
+      //   offeredCredentialConfiguration?.format === "mso_mdoc" &&
+      //   pidSchemes?.msoMdocNamespaces.includes(
+      //     offeredCredentialConfiguration.namespace
+      //   );
+      const shouldKeyBeHardwareBackedForSdJwtVc =
+        offeredCredentialConfiguration?.format === 'vc+sd-jwt' &&
+        pidSchemes?.sdJwtVcVcts.includes(offeredCredentialConfiguration.vct)
+
+      // TODO: add mso-mdoc config from above
+      const shouldKeyBeHardwareBacked = shouldKeyBeHardwareBackedForSdJwtVc ?? shouldKeyBeHardwareBackedForMsoMdoc
```
TimoGlastra, line 219: ?? will only go to the next one if the value is undefined, but shouldKeyBeHardwareBackedForSdJwtVC will always have a value. I think you mean || in this case?

TimoGlastra (approved):

PR #147, opened 2024-08-20
fix: add text encoder polyfill

berendsliedrecht:


TimoGlastra (approved):
Is this also why it crashed in prod?

berendsliedrecht:
> Is this also why it crashed in prod?

I believe so. This was the only error I could find in the logcat trace

PR #148, opened 2024-08-20
feat: wallet screen detail screen

TimoGlastra:
Adds stuff for the wallet screen and detail screen. Activity is stil todo.

@janrtvld some things maybe you could help with:
- showing toast in a modal screen (doesn't work atm -- needed for biometric failed toast)
- allowing headers to be transparent / not sohwn / hidden on scroll
  - i think on wallet screen we don't want a header?
  - on the pid detail screen we want to hide it on scroll maybe?
  - how to style the credential attributes rows background color differently in paradym vs ausweis wallet?

https://github.com/user-attachments/assets/c20f351c-2de9-4473-8162-ec650e3edd7a


janrtvld:
I'll take a look at it now.

janrtvld:
> showing toast in a modal screen (doesn't work atm -- needed for biometric failed toast)

Will look into this.

> i think on wallet screen we don't want a header?

Ideally we do it when you scroll, the header will scroll under the top navigation and a smaller version appears in the top navigation. Like this: https://imgur.com/a/84iq0BP


> on the pid detail screen we want to hide it on scroll maybe?

Hmm, I wouldn't bother for now. The current way has the back button always visible which I think is better UX. 

> how to style the credential attributes rows background color differently in paradym vs ausweis wallet?

Probably a 'theme' or 'variant' property on <CredentialAttributes /> which you use to style the TableContainer or TableRow with the correct colour.

janrtvld:
You can render toast inside a modal screen by rendering `<ToastViewport />` inside the screen component.

berendsliedrecht:
For after funke, we should probably check how we would like to display booleans. Because the age over thingy is quite weird unless you know the underlying data model.

TimoGlastra:
> For after funke, we should probably check how we would like to display booleans. Because the age over thingy is quite weird unless you know the underlying data model.

Worked, thanks!

PR #150, opened 2024-08-21
fix: b prime architecture proposal 2.0 changes

berendsliedrecht:


apps/ausweis/src/features/onboarding/onboardingContext.tsx
```diff
@@ -294,14 +295,21 @@ export function OnboardingContextProvider({
...
-      })
+    return (
+      secureUnlock
+        .setup(walletPin)
+        .then(({ walletKey }) => initializeAgent(walletKey))
+        // After `initializeAgent` function is finished we can assume that `setAgent(agent)` is called and the agent is set
+        // We store the wallet pin as the pid pin. We do this to avoid a double key derivation which is too much of a slow down
+        // In the future we can possibly sync the key derivation between what the
+        // Architecture Proposal suggests and what we require to do for our wallet storage
+        .then(deviceKeyPair.generate)
```
TimoGlastra, line 306: This should only be called in b' flow right?

apps/ausweis/src/use-cases/ReceivePidUseCaseBPrimeFlow.ts
```diff
@@ -210,36 +211,44 @@ export class ReceivePidUseCaseBPrimeFlow {
...
 
       this.accessToken = await acquireAccessToken({
         resolvedCredentialOffer: this.resolvedCredentialOffer,
         resolvedAuthorizationRequest: {
           ...this.resolvedAuthorizationRequest,
-          code: authorizationCode,
+          code,
         },
         agent: this.options.agent,
+        // @ts-ignore
```
TimoGlastra, line 236: why ts-ignore? acquireAccesstoken is from internal agent package right, so we should be able to type this correctly?

apps/ausweis/src/use-cases/ReceivePidUseCaseBPrimeFlow.ts
TimoGlastra, line 1: Need to make sure we keep these in sync, as there's a lot of overlap

packages/agent/src/display.ts
```diff
@@ -7,6 +7,8 @@ import { ClaimFormat, Hasher, JsonTransformer, SdJwtVcRecord } from '@credo-ts/c
 import { getHostNameFromUrl, sanitizeString } from '@package/utils'
 import { decodeSdJwtSync, getClaimsSync } from '@sd-jwt/decode'
 
+import type { SeedCredentialPidData } from '@ausweis/storage'
```
TimoGlastra, line 10: I think the dependencies are not great here. ausweis depends on agent, not the other way around. As we use a separate `usePidCredential` only available in ausweis wallet my assumption was that we will add all this SEED credential stuff only to the ausweis app. Is that OK?

packages/agent/src/display.ts
```diff
@@ -7,6 +7,8 @@ import { ClaimFormat, Hasher, JsonTransformer, SdJwtVcRecord } from '@credo-ts/c
...
+import { createFind } from 'rxjs/internal/operators/find'
```
TimoGlastra, line 11: not used

packages/agent/src/hooks/useCredentialsForDisplay.ts
```diff
@@ -1,27 +1,30 @@
 import { useMemo } from 'react'
 
+import { useSeedCredentialPidData } from '@ausweis/storage'
 import { getCredentialForDisplay } from '../display'
 import { useSdJwtVcRecords, useW3cCredentialRecords } from '../providers'
 
 export const useCredentialsForDisplay = () => {
   const { w3cCredentialRecords, isLoading: isLoadingW3c } = useW3cCredentialRecords()
   const { sdJwtVcRecords, isLoading: isLoadingSdJwt } = useSdJwtVcRecords()
+  const { isLoading: isLoadingSeedCredential, seedCredential } = useSeedCredentialPidData()
```
TimoGlastra, line 10: same same

packages/agent/src/invitation/handler.ts
```diff
@@ -115,14 +120,54 @@ export async function resolveOpenId4VciOffer({
...
+
+    const payload = jwt.payload
+
+    const toBeSigned = `${TypedArrayEncoder.toBase64URL(
+      TypedArrayEncoder.fromString(JSON.stringify(header))
+    )}.${TypedArrayEncoder.toBase64URL(TypedArrayEncoder.fromString(JSON.stringify(payload)))}`
+    const signature = await deviceKeyPair.sign(new Uint8Array(TypedArrayEncoder.fromString(toBeSigned)))
+    const jws = `${toBeSigned}.${TypedArrayEncoder.toBase64URL(signature)}`
+    return jws
+  }
```
TimoGlastra, line 158: is this b' specfiic? Should we prefix it with that? (just so we know what to remove once we remove this flow again)

packages/agent/src/providers/AgentProvider.tsx
```diff
@@ -15,7 +16,9 @@ export const AgentProvider = ({ agent, children }: PropsWithChildren<AgentProvid
   <NativeAgentProvider agent={agent}>
     <W3cCredentialRecordProvider agent={agent}>
       <SdJwtVcRecordProvider agent={agent}>
-        <ExchangeRecordDisplayMetadataProvider>{children}</ExchangeRecordDisplayMetadataProvider>
+        <SeedCredentialProvider agent={agent}>
```
TimoGlastra, line 19: would be really weird that paradym wallet imports @ausweis/storage to add seed credential provider. This needs to go in ausweis wallet

packages/agent/src/invitation/handler.ts
```diff
@@ -133,6 +176,8 @@ export async function acquireAccessToken({
       resolvedAuthorizationRequest,
       resolvedCredentialOffer,
       code: resolvedAuthorizationRequest.code,
+      dPopKeyJwk,
+      getCreateJwtCallback: getCreateJwtCallbackForBPrime,
```
TimoGlastra, line 180: I'm not sure we always need to provide this?

TimoGlastra:
I cleaned up some stuff which caused quite some merge conflicts in this PR. I resolved them and will now merge this PR, but you may have to do some additional fixed on #152 as well

PR #156, opened 2024-08-25
feat: B' presentation flow for Architecture Proposal 2.0

berendsliedrecht:
supersedes #152


apps/easypid/src/features/wallet/FunkeOpenIdPresentationNotificationScreen.tsx
```diff
@@ -41,11 +42,21 @@ export function FunkeOpenIdPresentationNotificationScreen() {
...
+        pidPin: '8376487',
+      })
+    })()
+      .then(() =>
+        getCredentialsForProofRequest({
+          agent,
+          data: params.data,
+          uri: params.uri,
+        })
+      )
```
TimoGlastra, line 59: weird formatting.

```suggestion
      requestSdJwtVcFromSeedCredential({
        agent,
        authorizationRequestUri: params.uri ?? 'TODO: this is temp anyways',
        pidPin: '8376487',
      })
      .then(() =>
        getCredentialsForProofRequest({
          agent,
          data: params.data,
          uri: params.uri,
        })
      )
```

Also re comment: `// TODO: we have to detect whether the verifier requests something from the b' flow (maybe check iss field in the pres def?)`

I think we can detect it based on whether the eph_key is included in the request (we should only include it if requesting b' flow on the RP side)

apps/easypid/src/features/wallet/FunkePidCredentialDetailScreen.tsx
```diff
@@ -17,6 +17,8 @@ export function FunkePidCredentialDetailScreen() {
     return null
   }
 
+  console.log(credential.attributes)
+
```
TimoGlastra, line 21:
```suggestion
```

apps/easypid/src/use-cases/ReceivePidUseCaseFlow.ts
```diff
@@ -45,8 +45,8 @@ export abstract class ReceivePidUseCaseFlow<ExtraOptions = {}> {
     return this.currentState
   }
 
-  protected errorCallbacks: AusweisAuthFlowOptions['onError'][] = [(e) => this.handleError(e)]
-  protected successCallbacks: AusweisAuthFlowOptions['onSuccess'][] = [
+  private errorCallbacks: AusweisAuthFlowOptions['onError'][] = [this.handleError]
+  private successCallbacks: AusweisAuthFlowOptions['onSuccess'][] = [
```
TimoGlastra, line 49: I think we need to have these as protected, as well we need it as `(e) => this.handleError(e)` to prevent incorrect this binding

packages/agent/src/invitation/handler.ts
```diff
@@ -155,6 +157,7 @@ export function getCreateJwtCallbackForBPrime() {
     )}.${TypedArrayEncoder.toBase64URL(TypedArrayEncoder.fromString(JSON.stringify(payload)))}`
     const signature = await deviceKeyPair.sign(new Uint8Array(TypedArrayEncoder.fromString(toBeSigned)))
     const jws = `${toBeSigned}.${TypedArrayEncoder.toBase64URL(signature)}`
+    console.log(jws)
```
TimoGlastra, line 160:
```suggestion
```

TimoGlastra (approved):

apps/easypid/src/features/wallet/FunkeOpenIdPresentationNotificationScreen.tsx
berendsliedrecht: Yeah that works

PR #159, opened 2024-08-26
feat: menu and activity row

janrtvld:
image


apps/easypid/app.config.js
```diff
@@ -30,7 +30,7 @@ const invitationSchemes = ['openid', 'openid-initiate-issuance', 'openid-credent
  * @type {import('@expo/config-types').ExpoConfig}
  */
 const config = {
-  name: `EasyPID ${variant.name}`,
+  name: `EasyPID${variant.name}`,
```
TimoGlastra, line 33: this annoyed me, nice

packages/ui/src/content/Icon.tsx
```diff
@@ -85,4 +90,10 @@ export const HeroIcons = {
   User: wrapHeroIcon(UserIcon),
   Bars3: wrapHeroIcon(Bars3Icon),
   EllipsisVertical: wrapHeroIcon(EllipsisVerticalIcon),
+  Activity: wrapHeroIcon(QueueListIcon),
+  Settings: wrapHeroIcon(Cog8ToothIcon),
+  Feedback: wrapHeroIcon(ChatBubbleBottomCenterTextIcon),
+  ChevronRight: wrapHeroIcon(ChevronRightIcon),
+  Menu: wrapHeroIcon(Bars3Icon),
+  Interaction: wrapHeroIcon(ArrowsRightLeftIcon),
```
TimoGlastra, line 98: I kept all the icon names the same as hero icons (so it's easy to find). But not sure if we want that. Just FYI that that pattern is now broken

TimoGlastra (approved):

PR #161, opened 2024-08-26
feat: store history

TimoGlastra:
adds activity records and shows the last three

will do the disclosed payloads in a follw up pr.

somehow the record updated event is not triggered when we add a activity, resulting in the main screen UI not updating... no clue why, it's weird. will look into it


packages/agent/src/invitation/handler.ts
```diff
@@ -251,7 +255,17 @@ export const receiveCredentialFromOpenId4VciOfferAuthenticatedChannel = async ({
     },
   })) as unknown as Array<string>
 
-  return credentials[0]
+  const credential = credentials[0]
+  // FIXME: not sure if the index of credentials will match?
```
janrtvld, line 259: did it?

janrtvld (approved):

packages/agent/src/invitation/handler.ts
TimoGlastra: 😍

TimoGlastra: we only request one, so not really needed to fix now

PR #166, opened 2024-08-27
feat: pin locked screen

berendsliedrecht:
- Jan will add a variant to the illustration container to give it a red color with the `variant="danger"`


janrtvld (approved):

apps/easypid/src/features/wallet/FunkeOpenIdPresentationNotificationScreen.tsx
```diff
@@ -150,6 +154,11 @@ export function FunkeOpenIdPresentationNotificationScreen() {
           return
         }
 
+        if (error instanceof PidIssuerPinLockedError) {
+          router.push('pinLocked')
```
TimoGlastra, line 158: Need to be careful with `push` as you can go back. I think it should maybe be router.replace as there's no way back

PR #174, opened 2024-09-04
feat: delete credential and ui improvements

janrtvld:
- Added header to Inbox screen
- Switched all sheets to Tamagui
- Added delete credential UI 

~~Still need to do the delete logic, but not sure how that works with all the credential types. Maybe you can point me in the right direction (or create a hook i can use :))? @TimoGlastra~~ It works!

https://github.com/user-attachments/assets/17aaeb42-dc3d-483e-961a-70e929294905

### DIFF
--- a/apps/paradym/package.json
+++ b/apps/paradym/package.json
@@ -10,6 +10,7 @@
 		"prebuild": "APP_VARIANT=development pnpm expo prebuild --no-install"
 	},
 	"dependencies": {
+		"@animo-id/expo-secure-environment": "*",
 		"@hyperledger/anoncreds-react-native": "*",
 		"@hyperledger/aries-askar-react-native": "*",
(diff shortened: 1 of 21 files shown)


apps/paradym/package.json
```diff
@@ -10,6 +10,7 @@
 		"prebuild": "APP_VARIANT=development pnpm expo prebuild --no-install"
 	},
 	"dependencies": {
+		"@animo-id/expo-secure-environment": "*",
```
janrtvld, line 13: I needed to add this to start the wallet, not sure why it's needed

TimoGlastra: What was the error you got otherwise?

TimoGlastra (approved):
Only would like to see the delete logic moved to the agent package

PR #175, opened 2024-09-12
feat: typography and optional pid

janrtvld:
I've restructured the typography so that it's more consistent. I've also made the PID optional during the onboarding. For now, it's disabled on the main screen and if you want to set it up you'll need to reset. I'm working on extracting the scanning logic out of the onboarding but that was a bit more work than i thought 🥲.

For now this removes the PID req so we can test easier with just regular EAA's. 

https://github.com/user-attachments/assets/8eb46b4b-c7fc-4825-91f0-4037c75cd35b


TimoGlastra:
Yeah the onboarding became quite complex sadly. How do you want to show it within the app on main screen, as a modal? Or a push screen? 

Would be nice if we can reuse the complete flow, as otherwise we'll have consistencies very quickly 😬

TimoGlastra (approved):

PR #176, opened 2024-09-13
feat: LSP interop testing improvements

TimoGlastra:
- Allow untrusted X509 certificates
- Allow skipping the eID card scanning with a simulated card when you enter app PIN '737837' and (PIN re-enter)  '276536'. This spells `SERVER` `BROKEN` if you follow the pin pad. I think we can also use this to get past apple review
- pass OID4VC conformance test suite (some additional fixes)
- Support `haip://` scheme


TimoGlastra:
Merging so I can submit to apple for review over the weekend 🤓

PR #180, opened 2024-09-26
feat: multi vp support

TimoGlastra:
Signed-off-by: Timo Glastra <timo@animo.id>

PR #181, opened 2024-10-02
feat: receiving and sharing qeaas

janrtvld:
https://github.com/user-attachments/assets/f28263c1-5e58-4526-9ec0-74a6662dac38


apps/easypid/src/app/(app)/_layout.tsx
```diff
@@ -52,60 +50,48 @@ export default function AppLayout() {
...
-              <Stack.Screen name="menu/feedback" options={headerNormalOptions} />
-              <Stack.Screen name="menu/settings" options={headerNormalOptions} />
-              <Stack.Screen name="menu/about" options={headerNormalOptions} />
-              <Stack.Screen name="activity/index" options={headerNormalOptions} />
-              <Stack.Screen name="activity/[id]" options={headerNormalOptions} />
-              <Stack.Screen name="pinConfirmation" options={headerNormalOptions} />
-              <Stack.Screen name="pinLocked" options={headerNormalOptions} />
-            </Stack>
-          </Reanimated.View>
+          <Stack key="1234" screenOptions={{ headerShown: false }}>
```
TimoGlastra, line 70: Key needed?

apps/easypid/src/features/receive/FunkeOpenIdCredentialNotificationScreen.tsx
```diff
@@ -0,0 +1,146 @@
...
+          step: 'verify-issuer',
+          progress: 33,
+          backIsCancel: true,
+          screen: (
+            <VerifyPartySlide
+              key="verify-issuer"
+              name={credential?.display.issuer.name}
+              logo={credential?.display.issuer.logo}
+              // @ts-expect-error issuer is not typed
+              domain={credentialRecord?.metadata.issuer as string}
```
TimoGlastra, line 116: We have the openid4vc metadata, we should use util methods to get / set this. As this will cause bugs. We need an unified way to store metadata for a cred

packages/agent/src/display.ts
```diff
@@ -452,3 +452,9 @@ export function getCredentialForDisplay(credentialRecord: W3cCredentialRecord |
     claimFormat: credentialRecord.credential.claimFormat,
   }
 }
+
+export const getReadableNameFromHost = (host: string) => {
+  const parts = host.split('.')
+  const domainName = parts.length >= 2 ? parts[parts.length - 2] : host
```
TimoGlastra, line 458: Feels dangerous. E.g. animosolutions.azurewebsites.com will result in azurewebsites i think?

TimoGlastra (approved):
Big pr so probably missed some stuff. And jist looked at code for now, haven't ran it yet

apps/easypid/src/features/receive/FunkeOpenIdCredentialNotificationScreen.tsx
janrtvld: I've removed the use of metadata values

packages/agent/src/display.ts
janrtvld: Also removed.

janrtvld:
@TimoGlastra added a video. I think it's easiest of you just run it and test it and lmk if you run into anything.

TimoGlastra:
Video looks nice! Thanks. I'm only not so sure about the share overlay with yellow top and then the shared attributes below on white. I think it should look a bit more like a card maybe? It's quite a big contrast. Can't we add an overlay with a certain transparency so white text on top of it is always visible (or black text)?

Also -- does it support multiple credentials being presented?

TimoGlastra:
Also -- we can do this later, but i think we might have to tweak the UI just a little bit so it's not exactly the NL wallet design (i think it's good inspiration, but also no need for it to look 100% the same i think?)

TimoGlastra:
Both can be improved later, this is already a very good improvements, so good to merge on my side 🙌

janrtvld:
> Video looks nice! Thanks. I'm only not so sure about the share overlay with yellow top and then the shared attributes below on white. I think it should look a bit more like a card maybe? It's quite a big contrast. Can't we add an overlay with a certain transparency so white text on top of it is always visible (or black text)?
> 
> 
> 
> Also -- does it support multiple credentials being presented? 

Can you share a screenshot of what you mean? I'm not following.

The UI supports sharing multiple credentials but there is an error thrown from a lower level so I've created a ticket out of it.

janrtvld:
> Also -- we can do this later, but i think we might have to tweak the UI just a little bit so it's not exactly the NL wallet design (i think it's good inspiration, but also no need for it to look 100% the same i think?)

Yes, this is in the back of my mind. I've already changed the view of attributes back to what we have in the Paradym wallet. This makes it stand out a bit more. Still thinking about a different view for the home cards as well.

TimoGlastra:
> The UI supports sharing multiple credentials but there is an error thrown from a lower level so I've created a ticket out of it.

That will be resolved by #180, but we need to wait for a new release of PEX

PR #182, opened 2024-10-08
chore: remove patches, support jarm

TimoGlastra:
- support jarm
- updates to use https://github.com/animo/aries-framework-javascript/tree/funke and https://github.com/sphereon-opensource/oid4vc/tree/funke
- remove mdoc (as it's too brittle at the moment). We will add it back in when ready


apps/easypid/src/crypto/bPrime.ts
```diff
@@ -422,6 +422,7 @@ export const requestSdJwtVcFromSeedCredential = async ({
 
     const [firstCredential] = credentialAndNotifications
     if (!firstCredential) throw new Error('Error retrieving credential.')
+    if (typeof firstCredential === 'string') throw new Error('string cred now allowed here')
```
berendsliedrecht, line 425: not allowed I assume?

patches/@credo-ts__openid4vc@0.5.11.patch
```diff
@@ -0,0 +1,24 @@
+diff --git a/build/shared/issuerMetadataUtils.js b/build/shared/issuerMetadataUtils.js
```
berendsliedrecht, line 1: do we want to keep this patch? Or do it in thefork?

berendsliedrecht (approved):

patches/@credo-ts__openid4vc@0.5.11.patch
TimoGlastra: as it's related to the removal of mdoc, i think we can keep it here for now (as the correct code is already in another branch). If it causes issues I'll move it

PR #187, opened 2024-10-14
feat: add option sheet and metadata

janrtvld:
https://github.com/user-attachments/assets/2ebe7c58-55b2-4db9-afe1-2d23e3307ead


janrtvld:
Alignd it differently so there is more space on the screen now. I'm still rethinking this screen. I don't think it's very clear the reason is supplied from the issuer. Also, the card structure is a bit too much like the NL wallet, but I do like it 😕. Ideas are welcome :).

IMG_BC5D368A6A5C-1

apps/easypid/src/features/share/FunkeRequestedAttributesDetailScreen.tsx
```diff
@@ -47,41 +75,90 @@ export function FunkeRequestedAttributesDetailScreen({
     return null
   }
 
+  const displayedMetadata = useMemo(() => {
```
TimoGlastra, line 78: can't we just not render undefined fields? this feels a bit redundant

packages/app/src/utils/formatCredentialSubject.ts
```diff
@@ -53,9 +53,6 @@ export function formatCredentialSubject(
 
   const isArray = Array.isArray(subject)
   for (const [key, value] of Object.entries(subject)) {
-    // omit id and type
-    if (key === 'id' || key === 'type') continue
```
TimoGlastra, line 57: this will restul in weird behaviour for w3c creds which rely on this

TimoGlastra (approved):

apps/easypid/src/features/share/FunkeRequestedAttributesDetailScreen.tsx
janrtvld: Hmm the formatDate is needed though

TimoGlastra: Shouldn't we do that in the getCredentialForDisplay?

janrtvld: 👍

PR #188, opened 2024-10-17
feat: updated credential detail

janrtvld:
A lot of UI stuff. Lifecycle stuff is still hardcoded and will be added later.

https://github.com/user-attachments/assets/ebcb89fc-e71b-4fd3-a536-80b2c367224d


apps/easypid/src/features/share/FunkeRequestedAttributesDetailScreen.tsx
```diff
@@ -52,12 +52,39 @@ export function FunkeRequestedAttributesDetailScreen({
...
+  const [elementPosition, setElementPosition] = useState(0)
+  const scrollViewRef = useRef<ScrollViewRefType>(null)
+
   useEffect(() => {
     navigation.setOptions({
       headerRight: () => <IconContainer icon={<HeroIcons.EllipsisHorizontal />} onPress={() => setIsSheetOpen(true)} />,
     })
   }, [navigation])
 
+  const toggleMetadataVisibility = () => {
```
TimoGlastra, line 64: looks quite similar to on the other page, is there a risk between them getting out of sync? It's ok like this, just a though

packages/app/src/components/DeleteCredentialSheet.tsx
```diff
@@ -0,0 +1,84 @@
...
+  id: CredentialForDisplayId
+  name: string
+}
+
+export function DeleteCredentialSheet({ isSheetOpen, setIsSheetOpen, id, name }: DeleteCredentialSheetProps) {
+  const { credential } = usePidCredential()
+  const toast = useToastController()
+  const { agent } = useAgent()
+  const [isLoading, setIsLoading] = useState(false)
+  const { withHaptics, error, success } = useHaptics()
```
TimoGlastra, line 21: I was kinda confused where the `success` and `error` methods came from. I think we should call them `successHaptics` and `errorHaptics`. Then it's immediatley clear

packages/app/src/components/DeleteCredentialSheet.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+  const onDeleteCredential = async () => {
+    if (credential && id === credential?.id) {
```
TimoGlastra, line 24: Do we use the same detail screen for both? I thought we had a different detila screen for the pid? In that case we can just not render the delete sheet right? I think this is OK, but it's a bit hacky as a very low-level component now does PID specific logic, while it would make more sense to just not render / or pass props from higher up that this credential can't be archived

packages/app/src/components/DualResponseButtons.tsx
```diff
@@ -8,6 +8,7 @@ interface DualResponseButtonProps {
   declineText?: string
   variant?: 'confirmation' | 'regular'
   align?: 'horizontal' | 'vertical'
+  removeBottomPadding?: boolean
```
TimoGlastra, line 11: I think `bottomPadding` with default value true would make more sense

TimoGlastra (approved):
Nice! Some nits but nothing blocking

apps/easypid/src/features/share/FunkeRequestedAttributesDetailScreen.tsx
janrtvld: Created a reusable hook for it.

TimoGlastra:
@janrtvld looking at the video you are not using `getSdJwtPidAttributesForDisplay` anymore? This formatted the PID credential specifically for display. E.g. now it shows `Jwk` part of holder, while the holder wasn't shown before.

I think we should show the jwk thumbprint there, but i think it would be good to keep the specific rendering for the pid credential. Maybe we should have a custom type rendering registry. If the `vct` / `doctype` matches a certain value, it will call a certain get attributes for display method. You can register these higher level.

TimoGlastra:
The `getSdJwtPidAttributesForDisplay` is defined in `usePidCredential`

janrtvld:
Yes i wanted to do it on a higher level so i don't have to worry about it with pid checks in the UI. Will look at it today.

janrtvld:
Ah i see i already did this in the follow up PR: https://github.com/animo/paradym-wallet/blob/ce6314c2896b553401fc4fff80aae2a6c114f758/apps/easypid/src/hooks/useCredentialsWithCustomDisplay.ts

Btw in this video JWK is only shown because I selected 'view metadata attributes'.

TimoGlastra:
> Btw in this video JWK is only shown because I selected 'view metadata attributes'. Do you still want to hide it even if users want to see the metadata?

Well it shouldn't be shown for the PID as the custom metadata filters it out. So it was mostly also to point out we aren't using the `Credential Information` anymore from the `usePidCredential`: https://github.com/animo/paradym-wallet/blob/52497eb8896443889b40c5cbb517d04e2ff47d93/apps/easypid/src/hooks/usePidCredential.tsx#L132-L146

I think we need to show it as a JWK thumbprint (so it's a string, not an object, it's not useful to show that to the user). 

How does the new metadata relate to the `getSdJwtPidAttributesForDisplay` method?

PR #193, opened 2024-10-30
fix: pid display fixes

janrtvld:


apps/easypid/src/app/(app)/activity/index.tsx
```diff
@@ -2,7 +2,7 @@ import { FunkeActivityScreen } from '@easypid/features/activity/FunkeActivityScr
 import { useLocalSearchParams } from 'expo-router'
 
 export default function Screen() {
-  const { host } = useLocalSearchParams()
+  const { did } = useLocalSearchParams()
```
TimoGlastra, line 5: we don't use dids, why did you change this?

TimoGlastra (changes requested):
not sure why the change to did. Maybe entityIdentifier? That could be a did or a https url in that case?

janrtvld:
I was testing with did because for paradym all the hostnames are agent.paradym.id. But this will be fixed anyway with openid fed later.

TimoGlastra (approved):

PR #194, opened 2024-10-30
feat: authorization code flow ui

janrtvld:
I've started with the authorization code flow UI. I'm not sure how it needs to be implemented so this is all guess work. I've made a separate flow because not sure how much specific logic there is, and also to keep the regular flow a bit cleaner. If it turns out it's very little difference we can merge them into the regular `<FunkeOpenIdCredentialNotificationScreen />`.

Maybe you can leave some feedback on things i'm missing or what I can still add to make it easier for you to implement.

https://github.com/user-attachments/assets/ec727cd5-26b0-48dc-a3d9-d8aa4fff8866

### DIFF
--- a/apps/easypid/package.json
+++ b/apps/easypid/package.json
@@ -46,6 +46,7 @@
 		"expo-status-bar": "~1.12.1",
 		"expo-system-ui": "~3.0.6",
 		"expo-updates": "~0.25.16",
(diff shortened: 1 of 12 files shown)


TimoGlastra:
This is a good start! I will integrate the flow in here and see what is still missing

TimoGlastra (approved):

PR #199, opened 2024-11-01
feat: add mdoc presentation

TimoGlastra:


apps/easypid/src/constants.ts
```diff
@@ -23,6 +23,7 @@ export const trustedX509Certificates = [
   animoFunkeRelyingPartyCertificate,
   ubiqueRootCertificate,
   oldAnimoFunkeRelyingPartyCertificate,
+  'MIIBQDCB6KADAgECAhB4Pa/TJGJ5OihZO+hyyBfpMAoGCCqGSM49BAMCMA0xCzAJBgNVBAYTAk5MMB4XDTcwMDEwMTAwMDAwMFoXDTI1MTEyMjA4MjIxMlowDTELMAkGA1UEBhMCTkwwOTATBgcqhkjOPQIBBggqhkjOPQMBBwMiAALcD1XzKepFxWMAOqV+ln1fybBt7DRO5CV0f9A6mRp2xaNKMEgwRgYDVR0RBD8wPYI7NDk3YS0yNDAyLTQwMDAtMTI1Ni02NTZkLTU0ZTktMjE1MC0zM2NmLTVkYzMubmdyb2stZnJlZS5hcHAwCgYIKoZIzj0EAwIDRwAwRAIgZBBWcc6NSnvVl3GXMyOw/7I/e93WjofuAuyEFNQ4Td4CIHeFFfwjsK81ZFew+fW1YaBnhob1WX4EREVTvRTCS4K4',
```
TimoGlastra, line 26:
```suggestion
```

apps/easypid/src/features/share/FunkeOpenIdPresentationNotificationScreen.tsx
```diff
@@ -102,10 +101,10 @@ export function FunkeOpenIdPresentationNotificationScreen() {
   const usePin = useMemo(() => {
     const isPidInSubmission =
       submission?.entries.some((entry) =>
-        entry.credentials.some((credential) => credential.id === pidCredential?.id)
```
TimoGlastra, line 105: @janrtvld i think the submission credentials should also just use the credentials for display. And I was thinking of adding a `isPid` to the credential for display. That way we can just access that

apps/easypid/src/hooks/useCredentialsWithCustomDisplay.ts
```diff
@@ -1,25 +1,51 @@
...
+
+    // No pids, just return
+    if (pidIndex === -1 || !pidCredential) return withoutPids
+
+    // Insert only one of the pids
+    return [
+      ...withoutPids.slice(0, pidIndex),
+      ...(includeHiddenCredentials
+        ? pidCredentials.map((p) => ({ ...p, isPid: true }))
+        : [{ ...pidCredential, isPid: true }]),
```
TimoGlastra, line 42: looks a bit hacky (and it is). But by default we don't include dupilcate PID, but if we are sharing the mdoc PID we do want to get it (so we trigger `includeHiddenCredentials`)

apps/easypid/src/hooks/usePidCredential.tsx
```diff
@@ -226,12 +211,9 @@ export function getPidMetadataAttributesForDisplay(
 
   if (claimFormat === ClaimFormat.MsoMdoc) {
     const { holder, issuedAt, validUntil, type, ...metadataRest } = metadata
-    const { issuing_authority, issuing_country, issuance_date, expiry_date, ...remainingAttributes } =
-      attributes as PidMdocAttributes
+    const { issuance_date, expiry_date } = attributes as PidMdocAttributes
 
     return {
-      issuing_authority,
-      issuing_country,
```
TimoGlastra, line 234: We need to rethink the metadata a bit. `issuing_country` and `issuing_authority` should be shows as that it will be shared I think? By adding them to metadata that is hidden unless you go to the overview page and then click show metadata. What do you think?

packages/app/src/components/CredentialAttributes.tsx
```diff
@@ -3,7 +3,8 @@ import { Heading, LucideIcons, Paragraph, TableContainer, TableRow, XStack, YSta
 import { formatCredentialSubject } from '../utils'
 
 type CredentialAttributesProps = {
-  subject: Record<string, unknown>
+  // biome-ignore lint/suspicious/noExplicitAny: <explanation>
+  subject: Record<string, any>
```
janrtvld, line 7: Doesn't `unknown` work?

apps/easypid/src/hooks/usePidCredential.tsx
TimoGlastra: Fair 👍  that's what i did now

packages/app/src/components/CredentialAttributes.tsx
TimoGlastra: no  and it doesn't really matter as it's just JSON. So i decided to go for the easy route

apps/easypid/src/hooks/useCredentialsWithCustomDisplay.ts
janrtvld: Is the mdoc PID different than the sd-jwt PID? because that is why I initially didn't think this was needed and we can just reuse the display for the sd-jwt one while under the hood the mdoc one is shared.

janrtvld: Maybe that is too much of an abstraction anyway so something like this had to happen

janrtvld (approved):

apps/easypid/src/hooks/useCredentialsWithCustomDisplay.ts
TimoGlastra: yes the structure is different :/ So it would require quite some abstraction ( and mapping between keys).

PR #201, opened 2024-11-04
feat: manual biometrics and background lock

janrtvld:


apps/easypid/src/app/authenticate.tsx
```diff
@@ -15,17 +23,19 @@ import { useResetWalletDevMenu } from '../utils/resetWallet'
...
   const isLoading =
     secureUnlock.state === 'acquired-wallet-key' || (secureUnlock.state === 'locked' && secureUnlock.isUnlocking)
 
+  // biome-ignore lint/correctness/useExhaustiveDependencies: canTryUnlockingUsingBiometrics not needed
   useEffect(() => {
     if (secureUnlock.state === 'locked' && secureUnlock.canTryUnlockingUsingBiometrics) {
       secureUnlock.tryUnlockingUsingBiometrics()
     }
-  }, [secureUnlock])
+  }, [])
```
TimoGlastra, line 38: This doesn't seem correct. What are you trying to do?

packages/app/src/provider/BackgroundLockProvider.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+  useEffect(() => {
+    const handleAppStateChange = (nextAppState: AppStateStatus) => {
+      if (nextAppState === 'background' || nextAppState === 'inactive') {
+        backgroundTimeRef.current = new Date()
+      } else if (nextAppState === 'active') {
+        if (backgroundTimeRef.current) {
+          const timeInBackground = new Date().getTime() - backgroundTimeRef.current.getTime()
+
+          if (timeInBackground > BACKGROUND_TIME_THRESHOLD && secureUnlock.state === 'unlocked') {
```
TimoGlastra, line 22: should we show a toast?

packages/ui/src/components/PinPad.tsx
```diff
@@ -71,15 +77,16 @@ const PinNumber = ({ character, onPressPinNumber, disabled }: PinNumberProps) =>
...
 }
 
-export const PinPad = ({ onPressPinNumber, disabled }: PinPadProps) => {
+export const PinPad = ({ onPressPinNumber, useBiometricsPad, disabled }: PinPadProps) => {
   const pinValues = [
     [PinValues.One, PinValues.Two, PinValues.Three],
     [PinValues.Four, PinValues.Five, PinValues.Six],
     [PinValues.Seven, PinValues.Eight, PinValues.Nine],
-    [PinValues.Empty, PinValues.Zero, PinValues.Backspace],
+    [useBiometricsPad ? PinValues.Biometrics : PinValues.Empty, PinValues.Zero, PinValues.Backspace],
```
TimoGlastra, line 89: there is https://oblador.github.io/react-native-keychain/docs/api/functions/getSupportedBiometryType.

Already used here: https://github.com/animo/paradym-wallet/blob/56e64ac227046c6b6d8906911ef0cdcf828182bc/packages/secure-store/secure-wallet-key/walletKeyStore.ts#L65

packages/app/src/provider/BackgroundLockProvider.tsx
```diff
@@ -0,0 +1,39 @@
...
+  const secureUnlock = useSecureUnlock()
+  const backgroundTimeRef = useRef<Date | null>(null)
+
+  useEffect(() => {
+    const handleAppStateChange = (nextAppState: AppStateStatus) => {
+      if (nextAppState === 'background' || nextAppState === 'inactive') {
+        backgroundTimeRef.current = new Date()
+      } else if (nextAppState === 'active') {
+        if (backgroundTimeRef.current) {
+          const timeInBackground = new Date().getTime() - backgroundTimeRef.current.getTime()
```
TimoGlastra, line 20: Shouldn't we already lock it in the background? 
So if th app becomes inactive /background we `setTimeout`. once it becomes active we clear timeout. if timeout is executed we lock the wallet

apps/easypid/src/app/authenticate.tsx
janrtvld: Currently, it prompts me for me biometrics right away, and if you cancel it prompts again until you reach 3.

Now it prompts once, if it fails or gets cancelled it stops, and you can prompt it again by pressing the biometrics button on the pinpad.

packages/app/src/provider/BackgroundLockProvider.tsx
janrtvld: Imo this is a pretty common security thing so not needed to explain it

apps/easypid/src/app/authenticate.tsx
TimoGlastra: Hmm okay. I see. We should still pass `secureUnlock.state` right?

TimoGlastra (approved):

TimoGlastra:
The biometric button didn't seem to work for me btw? Not sure why but pressing it didn't do anything

janrtvld:
Hmm ill keep an eye on it

TimoGlastra:
Yeah it really doesn't work on my end. The button does nothing

PR #202, opened 2024-11-05
feat(mdoc): proximity flow for Android functionality

berendsliedrecht:


apps/easypid/src/features/proximity/shareDeviceResponse.ts
```diff
@@ -0,0 +1,42 @@
...
+import type { EasyPIDAppAgent } from '@package/agent'
+import { DeviceRequest, DeviceResponse, MDoc, type MdocContext, parseIssuerSigned } from '@protokoll/mdoc-client'
+
+type ShareDeviceResponseOptions = {
+  sessionTranscript: Uint8Array
+  deviceRequest: Uint8Array
+  agent: EasyPIDAppAgent
+}
+
+export const shareDeviceResponse = async (options: ShareDeviceResponseOptions) => {
```
TimoGlastra, line 14: Is that it!??

apps/easypid/src/features/proximity/shareDeviceResponse.ts
```diff
@@ -0,0 +1,42 @@
...
+  const mdocs = await options.agent.mdoc.getAll()
+  const issuerSignedDocuments = mdocs.map((mdoc) => {
+    const docType = mdoc.getTag('DocType') as string
+    return parseIssuerSigned(TypedArrayEncoder.fromBase64(mdoc.base64Url), docType)
+  })
```
TimoGlastra, line 19: Can we do a pre-selection on the doctype? So we don't have to fetch all docs? I think the device request should contain some filters?

TimoGlastra (approved):
noice noice noice. Should we move it to a separate package as well? Most agent stuff is handled in agent now

PR #205, opened 2024-11-07
fix: unlocking with face id

janrtvld:
@TimoGlastra This fixes the unlocking issue, but it might break other things. Any reason why `setCanTryUnlockingUsingBiometrics` before unlocking?


TimoGlastra:
You can't try unlocking using bimoetrics, while trying to unlock using biometrics. That's why i set it to false. It's like setLoading(true) / setLoading(false) before /after an async call in a component

janrtvld:
But what is `setIsUnlocking` for then?

TimoGlastra:
I probably did it because the useEffect would otherwise retrigger? But as we now do it only once i think only setting it to false once we reached the max attempts makes sense

berendsliedrecht (approved):

PR #206, opened 2024-11-07
feat: lifecycle states

janrtvld:
- Added states on the credential detail page
- Added expired/revoked badges to credential card
- Added expired/revoked state to be used in the presentation screen when attributes from an expired/revoked card are requested.

Expiry and from date already works, the other states is just UI.

### DIFF
--- a/packages/agent/src/display.ts
+++ b/packages/agent/src/display.ts
@@ -368,6 +368,11 @@ export function filterAndMapSdJwtKeys(sdJwtVcPayload: Record<string, unknown>) {
       Object.entries(visibleProperties).map(([key, value]) => [key, recursivelyMapAttribues(value)])
     ),
     metadata: credentialMetadata,
+    raw: {
+      issuedAt: iat ? new Date(iat * 1000) : undefined,
+      validUntil: exp ? new Date(exp * 1000) : undefined,
+      validFrom: nbf ? new Date(nbf * 1000) : undefined,
+    },
   }
 }
 
@@ -417,6 +422,8 @@ export function getCredentialForDisplay(credentialRecord: W3cCredentialRecord | 
       attributes: mapped.visibleProperties,
       metadata: mapped.metadata,
       claimFormat: ClaimFormat.SdJwtVc,
+      validUntil: mapped.raw.validUntil,
+      validFrom: mapped.raw.validFrom,
     }
   }
   if (credentialRecord instanceof MdocRecord) {
@@ -484,6 +491,9 @@ export function getCredentialForDisplay(credentialRecord: W3cCredentialRecord | 
       validFrom: undefined,
     } satisfies CredentialMetadata,
     claimFormat: credentialRecord.credential.claimFormat,
+    validUntil: credentialRecord.credential.expirationDate
+      ? new Date(credentialRecord.credential.expirationDate)
+      : undefined,
   }
 }
 
--- a/packages/agent/src/display.ts
+++ b/packages/agent/src/display.ts
@@ -453,6 +453,8 @@ export function getCredentialForDisplay(credentialRecord: W3cCredentialRecord | 
         type: mdocInstance.docType,
       } satisfies CredentialMetadata,
       claimFormat: ClaimFormat.MsoMdoc,
+      validUntil: mdocInstance.validityInfo.validUntil,
+      validFrom: mdocInstance.validityInfo.validFrom,
     }
   }
 
@@ -494,6 +496,7 @@ export function getCredentialForDisplay(credentialRecord: W3cCredentialRecord | 
     validUntil: credentialRecord.credential.expirationDate
       ? new Date(credentialRecord.credential.expirationDate)
       : undefined,
+    validFrom: credentialRecord.credential.issuanceDate,
   }
 }
 
--- a/packages/agent/src/display.ts
+++ b/packages/agent/src/display.ts
@@ -496,7 +496,9 @@ export function getCredentialForDisplay(credentialRecord: W3cCredentialRecord | 
     validUntil: credentialRecord.credential.expirationDate
       ? new Date(credentialRecord.credential.expirationDate)
       : undefined,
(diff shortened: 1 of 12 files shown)


packages/agent/src/display.ts
```diff
@@ -484,6 +491,9 @@ export function getCredentialForDisplay(credentialRecord: W3cCredentialRecord |
       validFrom: undefined,
     } satisfies CredentialMetadata,
     claimFormat: credentialRecord.credential.claimFormat,
+    validUntil: credentialRecord.credential.expirationDate
```
TimoGlastra, line 494: No validFrom? I think we can use issuanceDate, as they renamed that to validFrom in v2 of w3c credential spec.

packages/agent/src/display.ts
```diff
@@ -484,6 +491,9 @@ export function getCredentialForDisplay(credentialRecord: W3cCredentialRecord |
...
+      ? new Date(credentialRecord.credential.expirationDate)
+      : undefined,
```
TimoGlastra, line 496: Also mdoc is missing. It had validity info as well

packages/app/src/components/CardInfoLifecycle.tsx
```diff
@@ -1,31 +1,192 @@
...
+  const onPressValidate = withHaptics(() => {
+    // Implement navigation to the setup eID card flow.
+    toast.show('Coming soon', { customData: { preset: 'warning' } })
+  })
+
+  if (validUntil || validFrom) {
+    return <CardInfoLimitedByDate validUntil={validUntil} validFrom={validFrom} />
+  }
+
+  return (
```
TimoGlastra, line 72: This will never render if validFrom/until is defined?

packages/app/src/components/CardInfoLifecycle.tsx
```diff
@@ -1,31 +1,192 @@
...
+    <>
+      {batchLeft && batchLeft <= 5 && (
+        <Stack pb="$4">
+          <Button.Solid bg="$grey-50" bw="$0.5" borderColor="$grey-200" color="$grey-900" onPress={onPressValidate}>
+            Refresh card <HeroIcons.ArrowPath ml="$-2" size={20} color="$grey-700" />
```
TimoGlastra, line 77: So i asked about refresh and they recommended me to use the C" issuance flow which does support refresh tokens. 

So the flow would be:
- authentication using eid card first time. Get batch and refresh token
- use refresh token in background to retrieve new token and get new batch
- if refresh token also expires we need to trigger eID card flow. 


But this will be a lot better already. So i think the "only x left in batch" should only be shown if we werent't able to automatically get new ones and re-authentication is required. But that's really a fallback.

We have the ReceivePidUseCaseFlow. Maybe if you can add these placeholder methods for now I'll implement these later:
- tryRetrieveBatchUsingRefreshToken


Then if that method fails because of unauthorized we need to trigger eID flow, or at least show to the user that they only have X verifications left and they need to re-authenticate.

packages/app/src/components/CardInfoLifecycle.tsx
```diff
@@ -1,31 +1,192 @@
...
+
+function CardInfoLimitedByDate({ validUntil, validFrom }: { validUntil?: Date; validFrom?: Date }) {
+  const [state, setState] = useState<CardInfoLimitedByDateState>('active')
   const [isOpen, setIsOpen] = useState(false)
   const { withHaptics } = useHaptics()
 
   const onPress = withHaptics(() => setIsOpen(!isOpen))
 
+  useEffect(() => {
+    if (validFrom && validFrom > new Date()) {
```
TimoGlastra, line 114: Can you compare dates like that? Don't you have to call getTime?

TimoGlastra: Also i'd handle expired before validFrom. If somehow it's validUntil is before validFrom

packages/app/src/components/CardWithAttributes.tsx
```diff
@@ -97,7 +105,9 @@ export function CardWithAttributes({
             </Heading>
           </YStack>
           <XStack h="$3">
-            {issuerImage?.url && <Image src={issuerImage.url} alt={issuerImage.altText} width={36} height={36} />}
+            {issuerImage?.url && !isRevokedOrExpired && (
```
TimoGlastra, line 108: Why don't we show the issuer image if revoked or expired?

packages/app/src/components/CardWithAttributes.tsx
```diff
@@ -39,6 +42,8 @@ export function CardWithAttributes({
   disclosedAttributes,
   disclosedPayload,
   disableNavigation = false,
+  isExpired = false,
```
TimoGlastra, line 45: I nowhere see these passed? Revocation is still todo, but this one we already know right?

Also for revocation we probably need to add a last checked time as well?

TimoGlastra (approved):
Nice!!

packages/app/src/components/CardInfoLifecycle.tsx
janrtvld: No, because if the many variants I moved that to a different component. (see line 69).

packages/app/src/components/CardInfoLifecycle.tsx
janrtvld: Yes it works like this.

I've moved the expired 👍

packages/app/src/components/CardWithAttributes.tsx
janrtvld: You still recognize the credential by it's background so it's not that big of a deal.

packages/app/src/components/CardWithAttributes.tsx
janrtvld: I've added the expired!

janrtvld: What do you mean with the last check time? Would you want to show that in the UI?

PR #208, opened 2024-11-12
feat: pid flow in wallet

janrtvld:
I tried to re-use some screens but i've copied over a lot of the logic because it's hard to understand code and some things just work a bit different if you already have a wallet. So to not make the logic in the onboarding even more unreadable i just did it like this.

TODOs
- [x] Test on Android

### DIFF
--- a/apps/easypid/src/features/pid/PidReviewRequestSlide.tsx
+++ b/apps/easypid/src/features/pid/PidReviewRequestSlide.tsx
@@ -1,8 +1,5 @@
-import { usePidDisplay } from '@easypid/hooks'
-import { Button, Circle, Heading, HeroIcons, Paragraph, Spinner, Stack, YStack } from '@package/ui'
-import { CardWithAttributes, DualResponseButtons, useWizard } from 'packages/app/src'
-import { sanitizeString } from 'packages/utils/src'
-import { useState } from 'react'
+import { Heading, YStack } from '@package/ui'
+import { useWizard } from 'packages/app/src'
 import { OnboardingIdCardRequestedAttributes } from '../onboarding/screens/id-card-requested-attributes'
 
 interface PidReviewRequestSlideProps {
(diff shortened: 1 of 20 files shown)


packages/ui/src/components/IdCard.tsx
```diff
@@ -103,7 +103,7 @@ export function IdCard({ icon, issuerImage, backgroundImage, userName, onPress,
             {onPress && <HeroIcons.ArrowRight color="$black" />}
           </XStack>
         </XStack>
-        <Stack zIndex="$-1" position="absolute" top={0} left={0} right={0} bottom={0}>
+        <Stack zIndex="$-1" position="absolute" bg="#F1F2F0" top={0} left={0} right={0} bottom={0}>
```
TimoGlastra, line 106: shouldn't this color come from the theme?

apps/easypid/src/features/pid/PidReviewRequestSlide.tsx
```diff
@@ -0,0 +1,37 @@
...
+  const { onNext } = useWizard()
+  const display = usePidDisplay()
+
+  const [isLoading, setIsLoading] = useState(false)
+
+  const onContinue = () => {
+    if (isLoading) return
+    setIsLoading(true)
+
+    // Do logic
```
TimoGlastra, line 23: what do you mean with do logic? shouldn't we do the logic here? isn't it implemented yet?

TimoGlastra (changes requested):
To me this duplication doesn't make sense, and the reason being the complexity makes less sense. I think we just made it 2x as complex by dupilcating the code. Everything now needs to be changed in two places. 

I think the correct way would be to create a better abstraction. So we can at least reuse the views  and logic, but still have control over what is different.

At least we should have the screens as separate components

apps/easypid/src/features/pid/PidReviewRequestSlide.tsx
janrtvld: This was placeholder, removed it.

TimoGlastra (approved):

PR #209, opened 2024-11-12
feat: app usable without local secure environment support

berendsliedrecht:
draft as we need a hosted version of the wallet service provider to make sure we can use it easier.

related PR: https://github.com/animo/funke-wallet-service-provider/pull/1


apps/easypid/.env
```diff
@@ -0,0 +1,2 @@
+EXPO_PUBLIC_WALLET_SERVICE_PROVIDER_AUTH_TOKEN=0dd5e0ac-f6a5-42ce-b945-09735c3a8f17
```
TimoGlastra, line 1: Auth token doesn't really work for mobile apps as you can extract it from the build.

Or is this temporary until we can do app integrity?

apps/easypid/src/crypto/WalletServiceProviderClient.ts
```diff
@@ -0,0 +1,52 @@
+import type { SecureEnvironment } from '@animo-id/expo-secure-environment'
+
+export class WalletServiceProviderClient implements SecureEnvironment {
```
TimoGlastra, line 3: Can we move this to a separate package?

apps/easypid/src/crypto/WalletServiceProviderClient.ts
```diff
@@ -0,0 +1,52 @@
...
+  public async generateKeypair(id: string): Promise<void> {
+    const response = await fetch(`${this.hsmUrl}/create-key`, {
+      method: 'POST',
+      headers: this.headers,
+      body: JSON.stringify({ keyType: 'P256', keyId: id }),
+    })
+    await response.text()
+  }
+
+  public async getPublicBytesForKeyId(keyId: string): Promise<Uint8Array> {
```
TimoGlastra, line 38: We will probably have to sign each request with a key or something for auth?

apps/easypid/src/crypto/WalletServiceProviderClient.ts
TimoGlastra, line 3: Also, i'm not really sure why this implements secure env? We're going yo have to interact a lot more with the wallet provider backend (for attestations etc). I think it makes more sense to create a separate service.

Also my understanding was we'd also create other keys in the HSM, but it now looks we only do the PID key?

TimoGlastra (commented):
Nice! Great first step, left some comments

apps/easypid/src/crypto/WalletServiceProviderClient.ts
TimoGlastra: Probably we should do: app integrity, issue access token bound to app integrity.

Also c' flow has some guidance on how to interact with hsm i think, did you look into that already?

apps/easypid/.env
berendsliedrecht: This was very temporary. Also, I should not push the .env file haha

apps/easypid/src/crypto/salt.ts
```diff
@@ -0,0 +1,25 @@
...
+
+  const crypto = new CredoWebCrypto(agent.context)
+
+  const saltBytes = crypto.getRandomValues(new Uint8Array(12))
+  const saltString = TypedArrayEncoder.toBase64URL(saltBytes)
+  await agent.genericRecords.save({ content: { salt: saltString }, id: GENERIC_RECORD_SALT_ID })
+  return saltString
+}
+
+const getSalt = async (agent: EasyPIDAppAgent): Promise<string | null> => {
```
TimoGlastra, line 18: a salt should be unique per key. So I think we should provide a salt id. or we should name these methods specific to the context of the salt (so e.g. WalletServiceProviderPinKeySalt)

TimoGlastra (approved):

apps/easypid/src/crypto/salt.ts
berendsliedrecht: `WalletServiceProviderPinKeySalt` thats fine. This will only be used for the device key, so renaming for the method would be good, yes!

PR #211, opened 2024-11-13
feat: ui flow for offline sharing

janrtvld:
There is one issue when you leave and come back again the QR code is sometimes invalid. I dont know if this is because i'm on a dev server or not, so we should see if this works when released. 

Here is a video of the flow (with fake data on iPhone, but i can't nicely record on Android).

https://github.com/user-attachments/assets/3f1f1cfa-da04-4f96-96e6-10eb3ab34a25


apps/easypid/src/app/(app)/notifications/offlinePresentation.tsx
```diff
@@ -0,0 +1,20 @@
...
+import { useLocalSearchParams } from 'expo-router'
+
+export default function Screen() {
+  const { sessionTranscript, deviceRequest, requestedAttributes } = useLocalSearchParams()
+
+  const sessionTranscriptArray = new Uint8Array(Buffer.from(sessionTranscript as string, 'base64'))
+  const deviceRequestArray = new Uint8Array(Buffer.from(deviceRequest as string, 'base64'))
+  const requestedAttributesArray = Array.isArray(requestedAttributes)
+    ? requestedAttributes
+    : JSON.parse(requestedAttributes as string)
```
TimoGlastra, line 11: I think the input will never be array right (with the json parse?)

```suggestion
  const { sessionTranscript, deviceRequest, requestedAttributes } = useLocalSearchParams<{ sessionTranscript: string; deviceRequest: string; requestedAttributes: string }>()

  const sessionTranscriptArray = new Uint8Array(Buffer.from(sessionTranscript, 'base64'))
  const deviceRequestArray = new Uint8Array(Buffer.from(deviceRequest, 'base64'))
  const requestedAttributesArray = JSON.parse(requestedAttributes)
```

apps/easypid/src/features/scan/FunkeQrScannerScreen.tsx
```diff
@@ -0,0 +1,283 @@
...
+          // Take the Doc item that matches with the mdoc pid type
+          // Only support one doc item for now
+
+          const requestedNamespace = data.requestedItems[0][pidSchemes.msoMdocDoctypes[0]]
+          if (!requestedNamespace) throw new Error('Unsupported credential requested.')
+
+          pushToOfflinePresentation({
+            sessionTranscript: Buffer.from(data.sessionTranscript).toString('base64'),
+            deviceRequest: Buffer.from(data.deviceRequest).toString('base64'),
+            requestedAttributes: Object.entries(requestedNamespace).map(([key]) => key),
```
TimoGlastra, line 237: do we only handle the requested namespaces (not attributes within?)

apps/easypid/src/features/scan/FunkeQrScannerScreen.tsx
```diff
@@ -0,0 +1,283 @@
...
+import easypidLogo from '../../../assets/easypid.png'
+import { checkMdocPermissions, getMdocQrCode, requestMdocPermissions, waitForDeviceRequest } from '../proximity'
+
+const unsupportedUrlPrefixes = ['_oob=']
+
+interface QrScannerScreenProps {
+  credentialDataHandlerOptions?: CredentialDataHandlerOptions
+}
+
+export function FunkeQrScannerScreen({ credentialDataHandlerOptions }: QrScannerScreenProps) {
```
TimoGlastra, line 34: it's sort of ok with the deadline now, but i think it would have been better to split this into a generic scan/show component that can be used in both paradym and easypid and then add a `showQrCode` and exrract all the mdoc specific logic out of here.

apps/easypid/src/features/share/slides/ShareCredentialsSlide.tsx
```diff
@@ -15,14 +15,19 @@ interface ShareCredentialsSlideProps {
   submission: FormattedSubmission
   onDecline: () => void
   isAccepting: boolean
+
+  verifierName?: string
```
TimoGlastra, line 19: better to make this `verifierDisplay` like we have issuer display which contains name / logo etc..

TimoGlastra (approved):

apps/easypid/src/features/share/slides/ShareCredentialsSlide.tsx
janrtvld: removed as this is currently empty anyway for offline flow. Will do that in the regular flow later.

apps/easypid/src/features/scan/FunkeQrScannerScreen.tsx
janrtvld: It takes the requestedNamespace which is equal to the pid, and in there are the attributes requested.

PR #212, opened 2024-11-18
feat: about screen and privacy policy

janrtvld:
Added text to about screen, and link to privacy policy. Also added temporary mailto link for the feedback item.

image


TimoGlastra:
the eye icon is not centered which looks a bit off

TimoGlastra (approved):

PR #214, opened 2024-11-19
chore: remove bprime

TimoGlastra:


package.json
```diff
@@ -28,14 +28,14 @@
...
 			"@hyperledger/indy-vdr-react-native": "^0.2.0",
-			"@credo-ts/anoncreds": "https://gitpkg.vercel.app/animo/aries-framework-javascript/packages/anoncreds?funke",
-			"@credo-ts/askar": "https://gitpkg.vercel.app/animo/aries-framework-javascript/packages/askar?funke",
-			"@credo-ts/cheqd": "https://gitpkg.vercel.app/animo/aries-framework-javascript/packages/cheqd?funke",
-			"@credo-ts/core": "https://gitpkg.vercel.app/animo/aries-framework-javascript/packages/core?funke",
-			"@credo-ts/indy-vdr": "https://gitpkg.vercel.app/animo/aries-framework-javascript/packages/indy-vdr?funke",
-			"@credo-ts/openid4vc": "https://gitpkg.vercel.app/animo/aries-framework-javascript/packages/openid4vc?funke",
-			"@credo-ts/question-answer": "https://gitpkg.vercel.app/animo/aries-framework-javascript/packages/question-answer?funke",
-			"@credo-ts/react-native": "https://gitpkg.vercel.app/animo/aries-framework-javascript/packages/react-native?funke",
+			"@credo-ts/anoncreds": "0.5.14-alpha-20241119114658",
```
berendsliedrecht, line 31: we are not using the custom versions anymore?

apps/easypid/src/crypto/bPrime.ts
```diff
@@ -407,6 +408,7 @@ export const requestSdJwtVcFromSeedCredential = async ({
         }
       },
       ...tokenResponse,
+      // @ts-ignore
```
berendsliedrecht, line 411: can't this be removed?

package.json
```diff
@@ -44,18 +44,10 @@
...
-			"@sphereon/oid4vci-issuer": "https://gitpkg.vercel.app/animo/OID4VC/packages/issuer?funke",
-			"@sphereon/oid4vci-client": "https://gitpkg.vercel.app/animo/OID4VC/packages/client?funke",
-			"@sphereon/jarm": "https://gitpkg.vercel.app/animo/OID4VC/packages/jarm?funke",
-			"@protokoll/mdoc-client": "0.2.36"
+			"react-native-reanimated": "~3.15.5"
 		},
 		"patchedDependencies": {
-			"@hyperledger/indy-vdr-react-native@0.2.2": "patches/@hyperledger__indy-vdr-react-native@0.2.2.patch",
-			"@protokoll/mdoc-client": "patches/@protokoll__mdoc-client.patch"
+			"@hyperledger/indy-vdr-react-native@0.2.2": "patches/@hyperledger__indy-vdr-react-native@0.2.2.patch"
```
berendsliedrecht, line 50: this patch is needed, otherwise Jan won't be able to do the flow

berendsliedrecht (changes requested):

package.json
TimoGlastra: i'll add them back in after -- but we don't have any custom code now. (we will after we add in federation). But i wanted to check types without bprime

apps/easypid/src/crypto/bPrime.ts
TimoGlastra: Yeah so you'd prefer removing all bprime code?

package.json
TimoGlastra: Yes see my question in slack. Can't we make a fix in the actual mdoc repo?

apps/easypid/src/crypto/bPrime.ts
berendsliedrecht: Ah no this is fine, no point actually in removing it :)

berendsliedrecht (approved):

PR #219, opened 2024-11-20
feat: display claim metadat

TimoGlastra:
Adds support for storing fetching sd-jwt type metadata and extracting the already supported display metadata from ti

Also adds support for storing the `claim` property for sd-jwt /mdoc for how to render claims

@janrtvld what we still need to do now is:
- if sd-jwt type metadata we need to extract the `claims` property and then for the path entries
  - see https://www.ietf.org/archive/id/draft-ietf-oauth-sd-jwt-vc-06.html#name-example-2-type-metadata
- if not -sdjwt type metadata we need to extract the `claims` and `order` property from the openid4vc metadata and use that for rendering. 
  - see https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#name-credential-issuer-metadata-6 and https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#name-credential-issuer-metadata-5

I think for the sd-jwt type metadata if there is no entry in the claim object, I'm not sure if we need to show the prop (or we should it under the metadata or something). I think we should also use that as the order.

So priority would be:
- use type metadata claims
- use oid4vci metadata order/claims
- use our default sanitizing logic


TimoGlastra:
I'll merge this as even without the claim metadata it adds support for basic sd-jwt type metadata

PR #220, opened 2024-11-20
fix: pid setup fix

janrtvld:
Not an ideal fix, still thinking how we can do it better.

The problem is that because we now do () => <Screen />, it re-renders everytime something changes in the properties passed to that screen. This means that once the state changes after the scan is complete, the screen renders again, and it triggers again. 

I tried fixing it with useMemo but that leads to the same thing. 

- [x] PIN screen still is a bit weird.


TimoGlastra:
Ah interesting. I think we can fix that by using keys maybe? It's now anonymous.

Reason i added it is because it would render screens before the required data was there.

But what if we just render conditionally? Then it won't render-render after defined

janrtvld:
It will still re-render if anything in the state changes that that screen depends on.

I allowed both for now, that works?

TimoGlastra (approved):

PR #224, opened 2024-11-22
feat: default to cloud hsm

berendsliedrecht:
Dependent on https://github.com/animo/expo-secure-environment/pull/18


TimoGlastra (approved):

berendsliedrecht:
> Let's first add the new secure env fix so it doesn't break the wallet

Yeah thats what I ment with the dependent on ... :)

berendsliedrecht:
Works when `setUSeCloudHsm` is on true and false.

PR #227, opened 2024-11-25
feat: dcql and a lot of UI stuff

TimoGlastra:
This PR grew a lot bigger than anticipated but it hasA LOT of great improvements:
- Add DCQL support  (some issues with showing disclosues, but flow is working and the logic is there)
- Remove all the custom pid display rendering. Instead it's now stored on the record as metadata
- Clean up a lot of the hacky display logic. It's way more unified now
- Pin working in all flows based on whether eID is used (todo in next pr: handle invalid pin, need to send 403 or some way to show in the wallet that the pin is invaild)
-


apps/easypid/src/agent/initialize.ts
```diff
@@ -30,7 +30,10 @@ export async function initializeAppAgent({
     await wsp.createSalt()
     await wsp.register()
   }
-  if (getShouldUseCloudHsm()) shouldUseFallbackSecureEnvironment(true)
+
+  const shouldUseCloudHsm = getShouldUseCloudHsm()
+  console.log('should use cloud hsm', shouldUseCloudHsm)
```
janrtvld, line 35:
```suggestion
```

apps/easypid/src/features/pid/WithBackPidRefresh.tsx
```diff
@@ -3,7 +3,7 @@ import { useBackgroundPidRefresh } from '../../hooks/useBackgroundPidRefresh'
 
 export function WithBackgroundPidRefresh({ children }: PropsWithChildren) {
   // Refresh PID once it reaches 1
-  useBackgroundPidRefresh(1)
+  // useBackgroundPidRefresh(1)
```
janrtvld, line 6: Is this broken?

apps/easypid/src/features/receive/FunkeCredentialNotificationScreen.tsx
```diff
@@ -132,6 +139,12 @@ export function FunkeCredentialNotificationScreen() {
     [agent]
   )
 
+  // TODO: Should we add this to the activitiy? We also don't do it for issuance
```
janrtvld, line 142: https://animosolutions.slack.com/archives/C0759K2BX0S/p1732446331079129?thread_ts=1730797262.934249&cid=C0759K2BX0S

apps/easypid/src/features/share/components/RequestedAttributesSection.tsx
```diff
@@ -14,71 +13,70 @@ export type RequestedAttributesSectionProps = {
...
+              </YStack>
+            )}
+            {unsatisfiedEntries.map((entry) => (
+              <CardWithAttributes
+                key={entry.inputDescriptorId}
+                name={entry.name ?? 'Credential'}
+                // We only have the attribute paths, no way to know how to render
+                // TODO: we could look at the vct?
+                // TODO: we should maybe support partial matches (i.e. vct matches), as then we can
+                // show a much better UI (you have the cred, but age is not valid, or this param is missing)
```
janrtvld, line 72: Do you mean then rendering the credential card but with 'date_of_birth' in red when it doesn't match the required age?

apps/easypid/src/use-cases/ReceivePidUseCaseCFlow.ts
```diff
@@ -67,7 +76,7 @@ export class ReceivePidUseCaseCFlow extends ReceivePidUseCaseFlow {
         resolvedCredentialOffer: this.resolvedCredentialOffer,
         credentialConfigurationIdsToRequest,
         clientId: ReceivePidUseCaseCFlow.CLIENT_ID,
-        requestBatch: getShouldUseCloudHsm() ? 10 : false,
+        requestBatch: getShouldUseCloudHsm() ? 2 : false,
```
janrtvld, line 79: Was this for testing?

apps/easypid/src/use-cases/RefreshPidUseCase.ts.ts
```diff
@@ -93,7 +94,7 @@ export class RefreshPidUseCase {
       resolvedCredentialOffer: this.resolvedCredentialOffer,
       credentialConfigurationIdsToRequest,
       clientId: RefreshPidUseCase.CLIENT_ID,
-      requestBatch: 10,
+      requestBatch: 2,
```
janrtvld, line 97: same here

janrtvld (approved):
nice

apps/easypid/src/features/pid/WithBackPidRefresh.tsx
TimoGlastra: Yeah for now. Will fix in a new pr

apps/easypid/src/features/receive/FunkeCredentialNotificationScreen.tsx
TimoGlastra: Okay. Reason for the comment is that i see this more as a failed issuance. But yeah if that's also on the list.

We don't support failed issuance though? And just adding this as failed proof request is weird i think as it's during issuance

apps/easypid/src/use-cases/ReceivePidUseCaseCFlow.ts
TimoGlastra: Need to update WSP. It doesn't batch soo is way too slow at 10

PR #253, opened 2024-12-02
some fixes and updates

TimoGlastra:
Three features / fixed in one PR 🙈 

- Added a development mode that shows internal error messages for easier debugging by LSPs [commit](https://github.com/animo/paradym-wallet/commit/a1aaf26655456082d15863d6f88edecfecaca598)
- Fixed an issue where the PIN screen would get stuck in a loading state when an incorrect PIN was entered  [commit](https://github.com/animo/paradym-wallet/commit/0f65ef98f5f26c3afc0968e4f848bf538a86cfd7)
- Fixed an issue with redirect based auth flow if the authorization flow left the in-app browser (e.g. when requiring authentication using the native AusweisApp with the eID card) [commit](https://github.com/animo/paradym-wallet/commit/eb333b81fe5662cc2f010e1ee9bbdc83a7e19aa3)

### DIFF
--- a/packages/app/src/components/SlideWizard.tsx
+++ b/packages/app/src/components/SlideWizard.tsx
@@ -165,7 +165,7 @@ export const SlideWizard = forwardRef(
               color={isError ? 'danger' : 'primary'}
             />
           </Stack>
-          <AnimatedStack style={animatedStyles} flex={1} jc="space-between">
+          <AnimatedStack style={animatedStyles} fg={1} pos="relative" jc="space-between">
             <ScrollableStack
               scrollViewProps={{
                 // @ts-expect-error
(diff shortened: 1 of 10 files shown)


packages/app/src/components/SlideWizard.tsx
```diff
@@ -165,7 +165,7 @@ export const SlideWizard = forwardRef(
               color={isError ? 'danger' : 'primary'}
             />
           </Stack>
-          <AnimatedStack pos="relative" style={animatedStyles} jc="space-between" fg={1}>
+          <AnimatedStack style={animatedStyles} flex={1} jc="space-between">
```
janrtvld, line 168: Is this intentional, and does this not break things on other screens?

janrtvld (approved):

packages/app/src/components/SlideWizard.tsx
TimoGlastra: Ah good question. this is of course not only used in this page. But yes it's intentional. If the content was longer than the screen the screen would overflow for the length of the component above. But will check if this doesn't break anything on the other screens.

PR #254, opened 2024-12-03
feat: experimental local ai model

janrtvld:
Performance is not that great yet. Will do some research on prompting techniques for smaller models, or we need to use the 3B model, that's also possible.

Still need to test on android, and decide on what devices we support. We can use expo-device to select the devices we want (can even filter on how much ram the device has)


apps/easypid/app.config.js
```diff
@@ -117,6 +117,9 @@ const config = {
       ],
     },
     associatedDomains: associatedDomains.map((host) => `applinks:${host}`),
+    entitlements: {
+      'com.apple.developer.kernel.increased-memory-limit': true,
```
TimoGlastra, line 121: any thing we also need to configure in appstoreconnect?

apps/easypid/src/features/menu/FunkeSettingsScreen.tsx
```diff
@@ -1,35 +1,46 @@
...
 
   return (
-    <FlexPage gap="$0" paddingHorizontal="$0">
-      <YStack w="100%" top={0} borderBottomWidth="$0.5" borderColor={isScrolledByOffset ? '$grey-200' : '$background'}>
-        <YStack gap="$4" p="$4">
-          <Stack h="$1" />
-          <Heading variant="h1" fontWeight="$bold">
-            Settings
-          </Heading>
+    <>
```
TimoGlastra, line 13: this is <> </> wrapper component is not used it seems?

apps/easypid/src/llm/useLLM.tsx
```diff
@@ -0,0 +1,182 @@
...
+    async (input: string): Promise<void> => {
+      if (!isModelReady || !isModelActivated) {
+        throw new Error('Model not active or still loading')
+      }
+      if (error) {
+        throw new Error(error)
+      }
+
+      try {
+        setResponse('') // This might be causing issues - let's move it
```
TimoGlastra, line 96: ?

TimoGlastra (approved):
very nice 👍

janrtvld:
Works on Android! It is however pretty slow, like a full minute before the result is in. I am using an older device though.

janrtvld:
Performance is better, but still a bit meh. I guess we can still run with it as it's an experimental feature anyway. I don't really want to use the 3B model as it's even slower then and we'd have to make the requirements much higher. 

It now works on devices with 4GB RAM and up, with higher end devices responding much quicker.

PR #260, opened 2024-12-20
fix: resolve issue with pnpm prebuild command

tusharbhayani:
### Summary 
- Resolved  issue with pnpm prebuild command


TimoGlastra (changes requested):
Hey! Could you please fix the formatting (you can run `pnpm style:fix` from the root)

TimoGlastra (approved):
Thanks :)

PR #262, opened 2025-01-13
feat: paradym wallet merge

janrtvld:
~~@TimoGlastra it's pretty simple but this was my idea for the flag setup. Lmk what you think.~~

Will use this as a temp branch to merge everything related to the PW/Funke merge.


apps/easypid/src/config/appType.ts
```diff
@@ -0,0 +1,36 @@
+import { APP_CONFIGS } from './features'
+
+export const appTypes = ['FUNKE_WALLET', 'PARADYM_WALLET'] as const
+export type AppType = (typeof appTypes)[number]
```
TimoGlastra, line 4: We can infer this based on the APP_CONFIGS

```ts
export type AppType = keyof typeof APP_CONFIGS
export const appTypes = Object.keys(APP_CONFIGS)
```

apps/easypid/src/config/features.ts
```diff
@@ -0,0 +1,19 @@
...
+  PARADYM_WALLET: {
+    EID_CARD: false,
+    AI_ANALYSIS: false,
+  },
+} satisfies Record<AppType, Record<FeatureKey, boolean>>
+
+export const FEATURES = {
+  EID_CARD: 'eid_card',
+  AI_ANALYSIS: 'ai_analysis',
+}
```
TimoGlastra, line 17: This is not really used? I think we can do an interface like this? 

```
interface Features {
  EID_CARD: boolean
  AI_ANALYSIS: boolean
}
```

TimoGlastra (approved):

TimoGlastra (approved):

PR #263, opened 2025-01-14
feat: pw onboarding

janrtvld:


apps/easypid/src/features/menu/FunkeAboutScreen.tsx
```diff
@@ -4,11 +4,22 @@ import React from 'react'
...
 import { Linking } from 'react-native'
 
+import { useIsFunkeWallet } from '@easypid/hooks/useFeatureFlag'
 import pj from '../../../package.json'
 
+const TEXT_FUNKE = `This app was created by Animo Solutions in the context of the SPRIN-D Funke ‘EUDI Wallet Prototypes’. It
+            serves as a prototype for future wallet providers. All code is available under Apache 2.0.`
+
+const TEXT_PARADYM =
+  'This app was created by Animo Solutions as a companion app for Paradym. All code is available under Apache 2.0.'
```
TimoGlastra, line 14: This is probably better handled in some sort of config file. That way we can just get it form the config file and it will be easier to update, add new wallets, etc.. 

Also thinking of translations going forward, there could be different translations files for this maybe?

It's ok for now, but probably good if you are going to do this in a lot of places to at some point extract it.

apps/easypid/src/features/pid/FunkePidSetupScreen.tsx
```diff
@@ -123,6 +125,17 @@ export function FunkePidSetupScreen() {
     onEnterPinRef.current.onEnterPin = onEnterPin
   }, [onEnterPin])
 
+  useEffect(() => {
```
TimoGlastra, line 128: This will never execute right?

TimoGlastra (approved):

PR #266, opened 2025-01-22
feat: didcomm in funke wallet

janrtvld:
It works, but I'm not that happy with the implementation yet. I need to find a way to combine SlideWizards, so a didcomm flow can have more than 1 screen.

- [ ] Add activity items for presentation failed/success
- [ ] DIDComm behind feature flag


apps/easypid/src/agent/index.ts
```diff
@@ -1,10 +1,38 @@
...
+import { useAgent } from '@package/agent'
+import { type EasyPIDAppAgent, type ParadymAppAgent, isEasyPIDAgent, isParadymAgent } from '@package/agent/src/agent'
 import { useSecureUnlock as _useSecureUnlock } from '@package/secure-store/secureUnlock'
+import type { initializeAppAgent } from './initialize'
 
 export { initializeAppAgent } from './initialize'
 
-export const useAppAgent = useAgent<EasyPIDAppAgent>
-export type AppAgent = EasyPIDAppAgent
+export type AppAgent = Awaited<ReturnType<typeof initializeAppAgent>>
```
janrtvld, line 8: What is the use of this second agent type?

janrtvld: Okay, but since we have only one app now, this will not really work right? Since we only know this during runtime, and not beforehand.

To fix this, I now made a second agent hook that is used for didcomm specific features, so you know you are using an agent that actually has didcomm enabled (see below).

packages/agent/src/hooks/useDidCommPresentationActions.ts
```diff
@@ -155,8 +155,7 @@ export function useDidCommPresentationActions(proofExchangeId: string) {
               return {
                 credential: credentialForDisplay,
                 disclosed: {
-                  // TODO: we don't show this yet on anoncreds screen, but we should add it
-                  attributes: {},
+                  attributes: credentialForDisplay.attributes,
```
TimoGlastra, line 158: This should only be the disclosed attributes?

packages/agent/src/format/formatPresentation.ts
```diff
@@ -121,6 +121,8 @@ export function formatDifPexCredentialsForRequest(
                   paths: getDisclosedAttributePathArrays(verifiableCredential.disclosedPayload, 2),
                 }
               } else {
+                console.log('hiero kom ik hier')
```
TimoGlastra, line 124: mooi!

apps/paradym/app/_layout.tsx
```diff
@@ -29,20 +27,19 @@ export const unstable_settings = {
...
   const { top } = useSafeAreaInsets()
-  const [isSettingUpMediation, setIsSettingUpMediation] = useState(false)
   useTransparentNavigationBar()
 
-  // Enable message pickup when mediation is configured and internet connection is available
-  useMessagePickup({
-    isEnabled: hasInternetConnection && isMediationConfigured,
-    agent,
+  // Only setup mediation if the agent is a paradym agent
+  useMediatorSetup({
```
TimoGlastra, line 39: This is totally ok for now, no need to spend more time on it, but this could also use the feature flags right? or is there a specific reason you didn't use it?

apps/easypid/src/features/share/slides/PinSlide.tsx
```diff
@@ -1,38 +1,28 @@
...
-
-        toast.show(r.result.title, {
-          message: r.result.message,
-          customData: { preset: 'danger' },
-        })
-
+    onPinComplete(
+      pin,
+      () => onNext(),
+      () => {
```
TimoGlastra, line 21: this is quite hard to understand if you don't know the specific order.

Maybe:?
```suggestion
      {
        pin,
        onComplete: () => onNext(),
        onError: () => {
```

apps/easypid/src/features/didcomm/useDidCommPresentationNotificationSlides.tsx
```diff
@@ -0,0 +1,104 @@
...
+          agent,
+          submission,
+          {
+            id: proofExchangeId,
+            name: verifierName,
+          },
+          'success'
+        )
+      })
+      .catch(async () => {
```
TimoGlastra, line 41: small chance, but if `addSharedActivityForSubmission` fails after successfull sharing, it will then call `addSharedActivityForSubmission` in catch again, and show the toast `Presentation could not be shared.`.

apps/easypid/src/features/didcomm/DidCommNotificationScreen.tsx
```diff
@@ -0,0 +1,129 @@
...
+  // We were routed here without an invitation
+  if (!params.invitation && !params.invitationUrl) {
+    // eslint-disable-next-line no-console
+    console.error('One of invitation or invitationUrl is required when navigating to DidCommNotificationScreen.')
+    pushToWallet()
+    return null
+  }
+
+  // Both flows have the same entry point, so we re-use the same loading request slide
+  // This way we avoid a double loading screen when the respective flow is entered
```
TimoGlastra, line 92: Nice 👍

apps/easypid/src/app/(app)/_layout.tsx
```diff
@@ -36,6 +47,16 @@ export default function AppLayout() {
     secureUnlock.state !== 'not-configured' && secureUnlock.state !== 'initializing' && !hasFinishedOnboarding
   const isWalletLocked = secureUnlock.state === 'locked' || secureUnlock.state === 'acquired-wallet-key'
 
+  // Only setup mediation if the agent is a paradym agent
+  useMediatorSetup({
```
TimoGlastra, line 51: I saw this hook also in another place?

apps/easypid/src/agent/initialize.ts
```diff
@@ -1,39 +1,52 @@
...
+        walletId: `easypid-wallet-${walletKeyVersion}`,
+        walletKey,
+        walletLabel: 'EasyPID Wallet',
+        trustedX509Certificates,
+      })
+    : await initializeParadymAgent({
+        keyDerivation: 'raw',
+        walletId: 'paradym-wallet-secure',
+        walletKey,
+        walletLabel: 'paradym-wallet',
```
TimoGlastra, line 27: Paradym Wallet?

TimoGlastra (approved):
Nice, good stuff 🙌

apps/easypid/src/agent/index.ts
TimoGlastra: Yes that makes sense. It should probably be the type of `EitherAgent` now (so you can only use feature available in both). Otherwise you need a specific hook. 

Could be e.g. `useDidcommAgent` even so it's more generic, but just `useParadymAgetn` and `useEasyPidAgent` are fine as well (it's temporary anyway I guess)

apps/paradym/babel.config.js
```diff
@@ -3,6 +3,7 @@ module.exports = (api) => {
   return {
```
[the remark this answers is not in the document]
janrtvld: Yeah was just temp to make it work again.

apps/paradym/app/_layout.tsx
janrtvld: No not really. I can add it to the check as well.

apps/easypid/src/app/(app)/_layout.tsx
janrtvld: It's also used in the old app.

packages/agent/src/hooks/useDidCommPresentationActions.ts
janrtvld: fixed

PR #267, opened 2025-01-22
feat(ios): proximity flow

berendsliedrecht:


apps/easypid/src/features/proximity/mdocProximity.ts
```diff
@@ -134,10 +133,12 @@ export const shareDeviceResponse = async (options: ShareDeviceResponseOptions) =
     .authenticateWithSignature(publicDeviceJwk, 'ES256')
     .sign(mdocContext)
 
+  console.log(Platform.OS, ':', Buffer.from(deviceResponse.encode()).toString('hex'))
```
TimoGlastra, line 136:
```suggestion
```

apps/easypid/src/features/proximity/mdocProximity.ts
```diff
@@ -134,10 +133,12 @@ export const shareDeviceResponse = async (options: ShareDeviceResponseOptions) =
...
+
   await mdt.sendDeviceResponse(deviceResponse.encode())
 }
 
 export const shutdownDataTransfer = () => {
-  const mdt = requireMdocDataTransfer().mdocDataTransfer.instance()
-  mdt.shutdown()
+  // const mdt = mdocDataTransfer.instance()
```
TimoGlastra, line 142: Is this broken?

apps/easypid/src/features/share/FunkeOpenIdPresentationNotificationScreen.tsx
```diff
@@ -218,7 +181,6 @@ export function FunkeOpenIdPresentationNotificationScreen() {
       trustedEntities={credentialsForRequest?.verifier.trustedEntities}
       lastInteractionDate={lastInteractionDate}
       onComplete={() => pushToWallet('replace')}
-      overAskingResponse={overAskingResponse}
```
TimoGlastra, line 221: Why remove this?

apps/easypid/src/features/wallet/FunkeOfflineQrScreen.tsx
```diff
@@ -126,12 +127,12 @@ export function FunkeOfflineQrScreen() {
     })
   )
 
-  useEffect(() => {
-    // Cleanup function that runs when component unmounts
-    return () => {
-      shutdownDataTransfer()
-    }
-  }, [])
+  // useEffect(() => {
```
TimoGlastra, line 130: Broken?

package.json
```diff
@@ -2,7 +2,6 @@
 	"name": "wallets",
 	"private": true,
 	"scripts": {
-		"postinstall": "pnpm deps:check",
```
TimoGlastra, line 5: ?

packages/agent/package.json
```diff
@@ -26,10 +25,19 @@
 	"peerDependencies": {
 		"@hyperledger/anoncreds-react-native": "catalog:",
 		"@hyperledger/aries-askar-react-native": "catalog:",
-		"@hyperledger/indy-vdr-react-native": "catalog:",
 		"expo": "catalog:"
 	},
 	"devDependencies": {
 		"@types/fast-text-encoding": "^1.0.3"
+	},
+	"expo": {
```
TimoGlastra, line 33: Do we need this in every package?

packages/agent/src/indyNetworks.ts
```diff
@@ -1,11 +1,11 @@
...
-    connectOnStartup: true,
-    isProduction: false,
-  },
-]
+// import type { IndyVdrPoolConfig } from '@credo-ts/indy-vdr'
```
TimoGlastra, line 1: File can be removed

patches/@hyperledger__anoncreds-react-native@0.2.4.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/ios/Anoncreds.mm b/ios/Anoncreds.mm
+index 6007a03d12a88cad8d64eedd10c9075ccaf4b84f..1a0a66c6e0e701c169aaa8476bf29fdae7a62495 100644
+--- a/ios/Anoncreds.mm
++++ b/ios/Anoncreds.mm
+@@ -3,7 +3,7 @@
+ #import <React/RCTUtils.h>
+ #import <ReactCommon/RCTTurboModule.h>
+ 
+-#import "turboModuleUtility.h"
++#import <anoncreds/turboModuleUtility.h>
```
TimoGlastra, line 10: Is this because of static? Can we add this to the library itself by detecting the mode with an #if or something (for later, but so we can use the libraries without patches)

TimoGlastra (changes requested):
Very nice!! Left some comments related to cleanup

apps/easypid/src/features/wallet/FunkeOfflineQrScreen.tsx
berendsliedrecht: No this is bugged, it shutsdown after we navigate, but we must keep it alive and shut it down later.

package.json
berendsliedrecht: I disable it every time as when I want to update a dep I have to do it in 5 places and running `pnpm add expo@52` for example does not work as it does not apply the changes to the package.json because it errors..

I will enable it again, but I find it too strict. It should error in CI, not local dev.

packages/agent/package.json
berendsliedrecht: every package that has the expo dependency I added it.

It should only be required for expo + one of the excluded dependencies. But this suffices imo. We can remove it when we move to expo 52.

patches/@hyperledger__anoncreds-react-native@0.2.4.patch
berendsliedrecht: I am not really sure why we had to change this. It will be removed in the new architecture as well.

apps/easypid/src/features/share/FunkeOpenIdPresentationNotificationScreen.tsx
berendsliedrecht: Left over from when ai model did not work together with proximity library

package.json
TimoGlastra: I think the new pnpm feature to specify catalogs can help here.

Consistent versions is a good thing. Doing it only in ci is fine (if you can make a PR?)

TimoGlastra (approved):

PR #268, opened 2025-01-28
fix: issued at for credential list

TimoGlastra:


janrtvld (approved):
lol took a while to notice haha

PR #272, opened 2025-02-04
fix: patches for french playground

TimoGlastra:
All the patches needed for french playground


TimoGlastra:
Feedback welcome, but releasing quickly so it can be used at the test event

PR #274, opened 2025-02-11
feat: specific app deployments

janrtvld:
I think this works? It works when building locally and changing the env.


.github/workflows/continuous-deployment.yaml
```diff
@@ -52,13 +54,13 @@ jobs:
...
         run: pnpm install
 
-      - name: 🚀 Build 
+      - name: 🚀 Build
         run: pnpm build
 
       - name: 🚀 Build internal preview
         if: ${{ github.event.inputs.channel == 'preview' }}
-        run: (cd apps/${{ github.event.inputs.app }} && eas build --non-interactive --platform ${{ github.event.inputs.platform }} --profile preview --no-wait)
+        run: eas build --non-interactive --platform ${{ github.event.inputs.platform }} --profile ${{ env.APP_PROFILE }}-preview --no-wait
```
TimoGlastra, line 62: can we skip the cd? I think we still need to run it from the `apps/easypid` folder right?

apps/easypid/app.config.js
```diff
@@ -1,180 +1,59 @@
...
-    name: ' (Preview)',
-    mediatorDid: 'did:web:mediator.paradym.id',
-  },
-  production: {
-    bundle: '',
-    name: '',
-    mediatorDid: 'did:web:mediator.paradym.id',
-  },
+const { version: easypidVersion } = require('./package.json')
+const { version: paradymVersion } = require('../paradym/package.json')
```
TimoGlastra, line 2: We'll get rid of the paradym app soon, so I wouldn't detect on it like this. Probably good to have a consistent version for both (and increasing one will increase the other, so we see it more as a codebase version). We can take the highest version of the two as a start now

apps/easypid/app.config.js
```diff
@@ -1,180 +1,59 @@
...
+  PARADYM_WALLET: createBaseConfig({
+    name: 'Paradym Wallet',
+    scheme: 'id.animo.paradym',
+    icon: './assets/paradym/icon.png',
+    adaptiveIcon: './assets/paradym/adaptive-icon.png',
+    splash: './assets/paradym/splash.png',
+    slug: 'paradym-wallet',
+    version: paradymVersion,
+    bundleId: 'id.paradym.wallet',
+    additionalInvitationSchemes: ['didcomm'],
```
TimoGlastra, line 35: We should probably also add `id.animo.paradym` here to allow for correct deeplinking

TimoGlastra: Also we need to extract it and use it here: https://github.com/animo/paradym-wallet/blob/e9b754122ba9e872de4e41e89d263a74157e1917/apps/easypid/src/features/receive/FunkeCredentialNotificationScreen.tsx#L53

TimoGlastra: Seems we can access it like this: https://docs.expo.dev/versions/latest/sdk/constants/#platformmanifest

apps/easypid/assets/funke/adaptive-icon.png
TimoGlastra, line 1: intentional that it's over the P here, and not in the other image?

apps/easypid/src/config/copy.ts
```diff
@@ -20,3 +20,10 @@ export const copy = {
 export function useAppCopy() {
   return copy[CURRENT_APP_TYPE]
 }
+
+export function useAppIcon() {
```
TimoGlastra, line 24: This is ok for now. But this is probably also something we can get from the app config at some pont

TimoGlastra (approved):

apps/easypid/assets/funke/adaptive-icon.png
janrtvld: Yeah adaptive icon often get's cropped into a circle, so it needs to be a bit more to the middle.

apps/easypid/src/config/copy.ts
janrtvld: How would you do this? Expo constants only returns the path from the root, like this:

`./assets/funke/icon.png`

But when I use it I need to use require, but that doesn't accept dynamic values. I also need relative paths.

How i use it now: 
```
            <Image height={96} width={96} src={require('../../../../assets/funke/icon.png')} />
```

this is not possible:
```
            <Image height={96} width={96} src={require(ExpoConstants.expoConfig?.icon)} />
```

TimoGlastra: Ah okay 👍

PR #294, opened 2025-02-20
fix: do not popup when secure enclave is not available

berendsliedrecht:


TimoGlastra:
I think we can revert this PR, as the google review used an old version? 

The popup was good for explaining

PR #297, opened 2025-02-24
fix: correctly handle pin flows

janrtvld:


apps/easypid/src/features/share/FunkeMdocOfflineSharingScreen.tsx
```diff
@@ -55,30 +57,34 @@ export function FunkeMdocOfflineSharingScreen({
...
-    }
-
-    setIsProcessing(true)
 
-    try {
-      await setWalletServiceProviderPin(pin.split('').map(Number))
-    } catch (e) {
-      setIsProcessing(false)
-      if (e instanceof InvalidPinError) {
+    if (shouldUsePin) {
```
TimoGlastra, line 64: Doest this also handle the biometric prompt canceled / incorrect error? 

There's two biometric errors. Not sure if they're handled somewhere higher/ lower. It would probably be good if we can create a proof sharing pipeline at some point, that will be used for ALL proof sharing, and you can toggle the options that are needed. As with the offline / openid /dc api / whatever it's hard to keep everything in sync


https://github.com/animo/paradym-wallet/blob/c0f3eef124f9b936e9b9da076cf32be85895e2d4/packages/app/src/features/notifications/OpenIdPresentationNotificationScreen.tsx#L76

TimoGlastra (approved):

apps/easypid/src/features/share/FunkeMdocOfflineSharingScreen.tsx
TimoGlastra: Yes secure env

janrtvld: Okay i handled it the same way as the other flows, but unable to test because secure env is broken.

PR #306, opened 2025-03-03
fix: minor improvs

janrtvld:
- Allow PID delete in dev mode
- Minor MDL custom view improvements

### DIFF
--- a/apps/easypid/src/features/wallet/FunkeCredentialDetailScreen.tsx
+++ b/apps/easypid/src/features/wallet/FunkeCredentialDetailScreen.tsx
@@ -15,6 +15,7 @@ import { useState } from 'react'
 import { useHeaderRightAction, useScrollViewPosition } from '@package/app/src/hooks'
 import { DeleteCredentialSheet, TextBackButton } from 'packages/app'
 
+import { useDevelopmentMode } from '@easypid/hooks'
 import { type CredentialForDisplayId, useCredentialForDisplayById } from '@package/agent'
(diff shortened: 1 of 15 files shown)


apps/easypid/src/features/wallet/FunkeCredentialDetailScreen.tsx
```diff
@@ -33,12 +34,13 @@ export function FunkeCredentialDetailScreen() {
...
 
   const { credential } = useCredentialForDisplayById(params.id)
+  const [isDevelopmentModeEnabled] = useDevelopmentMode()
   const [isSheetOpen, setIsSheetOpen] = useState(false)
 
   useHeaderRightAction({
     icon: <HeroIcons.Trash />,
     onPress: withHaptics(() => setIsSheetOpen(true)),
-    renderCondition: credential?.category?.canDeleteCredential ?? true,
+    renderCondition: (isDevelopmentModeEnabled || credential?.category?.canDeleteCredential) ?? true,
```
TimoGlastra, line 43: It's actually a requirement from ARF / EUDI to be able to remove the PID. So let's just enable it by default?

We should make sure to remove all instances though (so sd-jwt and mdoc)

TimoGlastra (approved):

apps/easypid/src/features/wallet/FunkeCredentialDetailScreen.tsx
janrtvld: alrighty

PR #311, opened 2025-03-05
feat: didcomm ui improvements

janrtvld:
- Added activity for new cards
- Added verifier screens for both didcomm issuance and verification
- Predicates are now shown in the requested attributes detail screen
- Purpose will be shown (once merged in Paradym)

### DIFF
--- a/apps/easypid/src/features/didcomm/useDidCommCredentialNotificationSlides.tsx
+++ b/apps/easypid/src/features/didcomm/useDidCommCredentialNotificationSlides.tsx
@@ -2,7 +2,9 @@ import { useParadymAgent } from '@easypid/agent'
 import { useDidCommCredentialActions } from '@package/agent'
 import type { SlideStep } from '@package/app/src'
 import { useToastController } from '@package/ui'
+import { addReceivedActivity, useActivities } from '../activity/activityRecord'
 import { CredentialRetrievalSlide } from '../receive/slides/CredentialRetrievalSlide'
+import { VerifyPartySlide } from '../receive/slides/VerifyPartySlide'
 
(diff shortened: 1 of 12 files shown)


apps/easypid/src/features/didcomm/useDidCommCredentialNotificationSlides.tsx
```diff
@@ -19,12 +21,20 @@ export function useDidCommCredentialNotificationSlides({
...
     await acceptCredential().catch(() => {
       toast.show('Something went wrong while storing the credential.', { customData: { preset: 'danger' } })
       onCancel()
     })
+    await addReceivedActivity(agent, {
+      entityId: credentialExchange?.connectionId as string,
+      name: display.issuer.name,
+      logo: display.issuer.logo,
+      backgroundColor: '#ffffff', // Default to a white background for now
+      credentialIds: [`w3c-credential-${credentialExchange?.id}`],
```
TimoGlastra, line 36: the exchange record is different from the crednetial. If we want the actual credential id we should do  `credentialExchange .credentials[0].credentialRecordId`

TimoGlastra (approved):

apps/easypid/src/features/didcomm/useDidCommCredentialNotificationSlides.tsx
TimoGlastra: @janrtvld i think this was not resolved?

janrtvld: @TimoGlastra hmm credentialExchange.credentials seems to be an empty array just after the credential is accepted. Could it be that it has not processed yet?

PR #312, opened 2025-03-06
Update app.config.js

janrtvld:
It's a nightmare to change the bundleId.


TimoGlastra (approved):
is this reverting back to the old one? I think we can't really change the id right?

janrtvld:
Yes it's a revert

PR #315, opened 2025-03-10
feat: initial qes flow

janrtvld:
- Error shown when having multiple transactions
- Error shown when transaction is not qes_authorization
- Activity events added for card signing and card added
- Show other cards in the request when requesting with a transaction
- search for matching credential in presentation instead of using the first one


apps/easypid/src/features/share/components/RequestedAttributesSection.tsx
```diff
@@ -18,66 +18,64 @@ export function RequestedAttributesSection({ submission }: RequestedAttributesSe
...
-              ? 'The following cards will be shared.'
-              : satisfiedEntries.length === 0
-                ? `You don't have the requested card(s).`
-                : `You don't have all of the requested cards.`}
-          </Paragraph>
-        </YStack>
-        {/* We always take the first one for now (no selection) */}
-        {satisfiedEntries.map(({ credentials: [credential], ...entry }) => {
-          return (
+      <YStack gap="$2">
```
janrtvld, line 21: this file is not changed, just removed the outer stack.

packages/agent/src/invitation/shareProof.ts
```diff
@@ -83,8 +86,16 @@ export const shareProof = async ({
             credentials: dcqlCredentials,
           }
         : undefined,
-      // FIXME: need to integrate with the QES PR, we should not just accept transaction data
-      // transactionData: resolvedRequest.transactionData?.map((e) => ({ credentialId: e.matchedCredentialIds[0] })),
+      transactionData:
+        resolvedRequest.transactionData && acceptTransactionData
+          ? (() => {
```
janrtvld, line 91: It still only uses dcql, is that correct? maybe you can take a look.

packages/agent/src/invitation/transactions.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export const getFormattedTransactionData = (credentialsForRequest?: CredentialsForProofRequest) => {
+  if (!credentialsForRequest) return undefined
+
+  const transactionData = credentialsForRequest.transactionData
+
+  if (!transactionData || transactionData.length === 0) return undefined
+
+  // Only allow one transaction data entry
+  if (transactionData.length < 1) throw new Error('Multiple transactions are not supported yet.')
```
TimoGlastra, line 14: I guess should be `>` instead of `<`?

```suggestion
  if (transactionData.length > 1) throw new Error('Multiple transactions are not supported yet.')
```

TimoGlastra (approved):

PR #317, opened 2025-03-12
feat: working inbox screen

janrtvld:


apps/easypid/src/features/didcomm/DidCommNotificationScreen.tsx
```diff
@@ -31,8 +35,43 @@ export function DidCommNotificationScreen() {
...
+    }
+  }
+
+  const onCancel = () => {
+    if (notification) {
+      if (notification.type === 'credentialExchange' || params.credentialExchangeId) {
+        void agent.modules.credentials.deleteById(notification.id ?? params.credentialExchangeId)
+      } else if (notification.type === 'proofExchange' || params.proofExchangeId) {
+        void agent.modules.proofs.deleteById(notification.id ?? params.proofExchangeId)
+      }
```
TimoGlastra, line 53: this cod  will only execute if there is a notification. In which case is notification defined, but notificadtion.id is undefined, and it will fallback to params?

Also, can't we populate `notification` based on the params? that way we can handle them in the same way here

apps/easypid/src/features/wallet/FunkeWalletScreen.tsx
```diff
@@ -47,8 +47,9 @@ export function FunkeWalletScreen() {
       </YStack>
 
       <FlexPage safeArea="y" fg={1} flex-1={false} bg="transparent">
-        <XStack pt="$2">
+        <XStack pt="$2" jc="space-between">
           <IconContainer bg="white" aria-label="Menu" icon={<HeroIcons.Menu />} onPress={pushToMenu} />
+          {isParadymWallet() && <InboxIcon />}
```
TimoGlastra, line 52: This should use feature flag instead. If didcomm is supported we want to show the inbox (it will become easier if we have another variant of the wallet later)

PR #318, opened 2025-03-18
feat: cache images

janrtvld:
I think this also fixes the image issue we've been having.


apps/easypid/src/features/receive/slides/CredentialCardSlide.tsx
```diff
@@ -1,6 +1,6 @@
 import type { CredentialDisplay } from '@package/agent'
 import { Heading, MiniCardRowItem, Paragraph, Stack, YStack } from '@package/ui'
-import { DualResponseButtons, useHasInternetConnection, useWizard } from 'packages/app/src'
+import { DualResponseButtons, useWizard } from 'packages/app/src'
```
TimoGlastra, line 3:
```suggestion
import { DualResponseButtons, useWizard } from '@package/app'
```

TimoGlastra (changes requested):
I don't think this will fix the issue with the local images.

The problem is:
- we import e.g. diriving privileges car image -> resolved to number 11
- we store this in the credential metadata, so the credential metadata now referes to file 11
- we change the app assets and add a new image. Number 11 now resolves to a different image
- you open the wallet, image 11 is resolved.

So we need to do either:
- do not store the paths on the credential metadata but replace them dynamically (not sure how to easily do this)
- actually store the path of the image. The import in expo/RN just resolves to a number.

TimoGlastra:
See this guide: https://dev.to/fdefreitas/how-to-obtain-a-uri-for-an-image-asset-in-react-native-with-expo-7bm

TimoGlastra:
Might be interesting to experiment with blurhash as well at some point: https://github.com/mrousavy/react-native-blurhash. Might give a cool effect for backgrounds etc.

janrtvld:
Expo-image also supports blurhashes, we just need a way to generate them.

TimoGlastra:
> Expo-image also supports blurhashes, we just need a way to generate them. 

Only for ios it seems

TimoGlastra:
> Maybe this one is even better: https://docs.expo.dev/develop/user-interface/assets/#load-an-asset-at-build-time-with-expo-asset-config-plugin. We can just link the assets and use a static local URI.
> 
> 

Downside seems to be you need to provide a width/height:
> Note that when rendering assets without using require, you need to explicitly provide a width / height.

janrtvld:
I'll pick up the local images issue in a follow up PR

PR #319, opened 2025-03-18
chore: wallet migration

TimoGlastra:
It also revamps the wallet reset (Again)., but i think it now works relatively solid.

The flow for the mgiration basically just shows the paradym onboarding, but once you're through that you will have your original credentials. I think that is good enough behaviour


janrtvld:
Anyway this can be tested locally?

TimoGlastra:
> > The flow for the mgiration basically just shows the paradym onboarding, but once you're through that you will have your original credentials. I think that is good enough behaviour
> 
> 
> 
> So you don't notice anything basically? Existing wallets will start on the new onboarding flow just like new wallets but will then have their original credentials.

Yeah. Should we add a new screen at the start mentioning the wallet will be migrated? I wanted to keep it simple, the PW is mostly for demo purposes anyway atm

TimoGlastra:
Merging for now, we can still add the screen

PR #321, opened 2025-04-01
feat: allow didcomm connection without issue/verify

TimoGlastra:
Made a start. It's a bit more work than i thought due to how the notification inbox works.

Basically how i want to spilt it:
- parse oob invitation -> get back the flow (issue/connect/veirfy) and whether a connection will be created, and whether a connection already exists
- based on th parsed oob invitation we show the different slides (not split them up)
- each flow handles the implementation. This will also call `acceptOutOfBandInvitation` and only then we have the credential/proof exchange record.


janrtvld:
I have a messy version working [here](https://github.com/animo/paradym-wallet/tree/feat/allow-connections), but running into some issues with the hooks (i think you had the same). So need to do some refactoring.

Ideally yes we split it up, but then you have a double LoadingSlide which looks weird, so that is why i initially went with one flow and dynamically add the slides. However this is becoming very messy so need to find a good structure to handle this.

PR #322, opened 2025-04-02
fix: use expo local assets

janrtvld:
- [ ] need to fix the credential background, it looks different and i dont know why


TimoGlastra:
>  need to fix the credential background, it looks different and i dont know why

Is this still todo? OR did you fix it?

janrtvld:
Fixed!

PR #324, opened 2025-04-07
feat: allow didcomm connection without issue/verify

janrtvld:
Took some time because I needed to update how SlideWizard handles this. We can use multiple SlideWizards together for 1 flow. This way we can dynamically build up the wizard based on the flow logic.


apps/easypid/src/features/didcomm/ConnectionSlides.tsx
```diff
@@ -0,0 +1,79 @@
...
+        <YStack gap="$4" ai="center">
+          <Heading>Connection established</Heading>
+          <Paragraph ta="center">
+            You can now receive notifications from{' '}
+            {name && (
+              <>
+                {' '}
+                <Paragraph fontWeight="$semiBold">{name}</Paragraph>
+              </>
+            )}
```
TimoGlastra, line 66: If there is no name it will render "You can now receive notifications from" right?

apps/easypid/src/features/didcomm/DidCommNotificationScreen.tsx
```diff
@@ -113,50 +120,76 @@ export function DidCommNotificationScreen() {
...
+          progress: 33,
+          screen: <LoadingRequestSlide key="loading-request" isLoading={!resolvedInvitation} isError={!!errorReason} />,
+        },
+        {
+          step: 'verify-issuer',
+          progress: resolvedInvitation?.flowType === 'connect' ? 66 : 50,
+          screen: (
+            <VerifyPartySlide
+              key="verify-issuer"
+              type="connect"
```
TimoGlastra, line 178: Shouldn't this change based onthe invitation type? So it says receive card from, etc..?

apps/easypid/src/features/didcomm/DidCommNotificationScreen.tsx
```diff
@@ -113,50 +120,76 @@ export function DidCommNotificationScreen() {
...
+
+  if (flow?.type === 'connect' && readyToNavigate) {
+    return <ConnectionSlides name={display.connection.name} onCancel={onCancel} onComplete={onComplete} />
+  }
+
+  if ((flow?.type === 'issue' && readyToNavigate) || params.credentialExchangeId) {
+    const isExistingExchange = !!params.credentialExchangeId
+    const id = isExistingExchange ? params.credentialExchangeId : flow?.id
+
+    return (
```
TimoGlastra, line 140: Eventually it would be nice that if you're navigating here from an existing notification we still show the party you're interacting with? But that's definitely for a later iteration

packages/agent/src/hooks/useDidCommConnectionActions.ts
```diff
@@ -0,0 +1,57 @@
...
+  outOfBandInvitation: {
+    id: 'placeholder',
+    label: 'placeholder',
+    imageUrl: 'https://example.com/logo.png',
+  } as OutOfBandInvitation,
+  existingConnection: {
+    id: 'placeholder',
+    alias: 'placeholder',
+    theirLabel: 'placeholder',
+    imageUrl: 'https://example.com/logo.png',
```
TimoGlastra, line 17: I guess these are all optional? so no need to set it?

apps/easypid/src/features/didcomm/DidCommNotificationScreen.tsx
janrtvld: Maybe we can show that in the notification list item?

PR #327, opened 2025-04-09
feat: update expo

janrtvld:


apps/easypid/base.app.config.js
```diff
@@ -130,7 +130,6 @@ const createBaseConfig = (appSpecific) => {
             compileSdkVersion: 35,
             useLegacyPackaging: true,
             extraMavenRepos: ['https://s01.oss.sonatype.org/content/repositories/snapshots/'],
-            kotlinVersion: '2.0.21',
```
TimoGlastra, line 133: Expo 52 uses Kotlin 2 by default right? Have you tested the DC API, as this needed kotlin 2.0

janrtvld: works!

janrtvld:
Okay i think i got most of it. Maybe we can do a test run with the preview release to catch some final issues

PR #332, opened 2025-04-15
feat: show which attributes are not present

janrtvld:
Do you think we should implement this lower, on the submission level? That code is very confusing to me so I did it like this but it might be better.

IMG_4449
<br/>
IMG_4450


TimoGlastra:
I think this appraoch won't really work for all types of credentials. To do it properly we need to address it in DCQL directly. Let's keep this open for a bit and we can merge it with the DCQL changes once ready

PR #335, opened 2025-04-16
fix: android status bars

janrtvld:


apps/easypid/src/app/(app)/(home)/scan.tsx
```diff
@@ -1,6 +1,21 @@
...
+import { QrScannerScreen, isAndroid } from '@package/app'
+import * as NavigationBar from 'expo-navigation-bar'
+import { useFocusEffect } from 'expo-router'
+import { useCallback } from 'react'
 import { credentialDataHandlerOptions } from '../_layout'
 
 export default function Screen() {
+  // make the navigation bar light when the camera scanner is open
+  useFocusEffect(
+    useCallback(() => {
```
TimoGlastra, line 10: Why do we need useCallback here? It can be the method directly right?

TimoGlastra (approved):

PR #337, opened 2025-04-22
fix: backgroundlock path state

janrtvld:
FUN-513

This doesn't work for requests right? because we can't receive a request again, that leads to an error.

Any way around that?

### DIFF
--- a/packages/app/src/provider/BackgroundLockProvider.tsx
+++ b/packages/app/src/provider/BackgroundLockProvider.tsx
@@ -1,40 +1,29 @@
-import { TypedArrayEncoder } from '@credo-ts/core'
-import { useSecureUnlock } from '@package/secure-store/secure-wallet-key/SecureUnlockProvider'
-import { usePathname } from 'expo-router'
+import type { PropsWithChildren } from 'react'
+
+import { useSecureUnlock } from '@easypid/agent'
 import { useRouter } from 'expo-router'
-import { type PropsWithChildren, useEffect, useRef } from 'react'
+import { useEffect, useRef } from 'react'
 import { AppState, type AppStateStatus } from 'react-native'
 
-const BACKGROUND_TIME_THRESHOLD = 60000 // 1 minute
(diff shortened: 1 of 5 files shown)


packages/app/src/provider/BackgroundLockProvider.tsx
```diff
@@ -1,29 +1,40 @@
...
           if (timeInBackground > BACKGROUND_TIME_THRESHOLD && secureUnlock.state === 'unlocked') {
             console.log('App was in background for more than 30 seconds, locking')
+            const lastPath = lastPathRef.current
             secureUnlock.lock()
-            router.replace('/authenticate')
+
+            if (lastPath && !['/', '/authenticate'].includes(lastPath)) {
+              // Encode the path for redirection after unlock
+              const encodedRedirect = TypedArrayEncoder.toBase64URL(TypedArrayEncoder.fromString(lastPath))
+              router.replace(`/authenticate?redirectAfterUnlock=${encodedRedirect}`)
```
TimoGlastra, line 34: I think the thing we want to fix is to handle a deeplinkg, but the wallet is closing the wallet on becoming active, so the deeplink isn't handled.

This may (not sure) fix the issue of a deeplink, but in case you were in a flow and went to the background, this won't work and require us to refactor a lot of of the logic, and will lead to bugs 🤔 

So i think we need to know here whether we are handling a deeplink or not. if not, we should just go to authenticate, if we do, we need to extract the deeplink url and redirect to that aftewards

janrtvld: For the flow bug with the background lock; you could handle inactive differently than background. E.g. when the modal opens to do the authorization code flow we have a state of inactive, and we could allow for a longer time (e.g. 2-3 minutes) then when you manually close the app into the background (1 minute).

TimoGlastra (approved):

TimoGlastra:
Let's wait with merging, as I think this will break the mDL flow with the AusweisApp (since that requires native deeplinking midflow)

PR #338, opened 2025-04-22
feat: dc api ui improvements

TimoGlastra:
Screenshot_20250422_183629_Google Play services

### DIFF
--- a/packages/agent/src/openid4vc/registerDcApi.ts
+++ b/packages/agent/src/openid4vc/registerDcApi.ts
@@ -1,5 +1,5 @@
 import { type RegisterCredentialsOptions, registerCredentials } from '@animo-id/expo-digital-credentials-api'
-import { DateOnly, type MdocNameSpaces } from '@credo-ts/core'
+import { DateOnly, type Logger, type MdocNameSpaces } from '@credo-ts/core'
 import { sanitizeString } from '@package/utils'
 import * as ExpoAsset from 'expo-asset'
 import { Image } from 'expo-image'
@@ -70,7 +70,7 @@ function mapSdJwtAttributesToClaimDisplay(claims: object, path: string[] = []): 
 /**
  * Returns base64 data url
  */
-async function resizeImageWithAspectRatio(asset: ExpoAsset.Asset) {
+async function resizeImageWithAspectRatio(logger: Logger, asset: ExpoAsset.Asset) {
   try {
     // Make sure the asset is loaded
     if (!asset.localUri) {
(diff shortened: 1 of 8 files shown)


TimoGlastra:
This has some issues when large images are used. Need to dig into it a bit more. Maybe it's the React Native bridge and we need to load the images in Kotlin, or we need compress them or something?

janrtvld:
Large as in image aspect ratio or large as in the size of the image file? & what is the issue? Slow?

packages/agent/src/openid4vc/registerDcApi.ts
```diff
@@ -64,16 +67,40 @@ function mapSdJwtAttributesToClaimDisplay(claims: object, path: string[] = []):
   })
 }
 
+async function loadCachedImageAsBase64DataUrl(url: string) {
+  const path =
+    // in case of external iamge
```
janrtvld, line 72:
```suggestion
    // in case of external image
```

packages/agent/src/openid4vc/registerDcApi.ts
```diff
@@ -105,15 +139,18 @@ export async function registerCredentialsForDcApi(agent: EitherAgent) {
...
 
-  const credentials = [...sdJwtCredentials, ...mdocCredentials]
+  const credentials = await Promise.all([...sdJwtCredentials, ...mdocCredentials])
   agent.config.logger.trace('Registering credentials for Digital Credentials API', {
     credentials,
   })
+
   await registerCredentials({
     credentials,
+    matcher: 'ubique',
```
janrtvld, line 154: Testing?

janrtvld (approved):

TimoGlastra:
> Large as in image aspect ratio or large as in the size of the image file? & what is the issue? Slow?

It just get's stuck, but the image rendered is only 20x20 so i  am going to look at ExpoImageManipulator and resize all images.

packages/agent/src/openid4vc/registerDcApi.ts
TimoGlastra: The library supports multiple matchers now and we use this one in our wallet, so not for testing

packages/agent/src/openid4vc/registerDcApi.ts
```diff
@@ -64,16 +67,95 @@ function mapSdJwtAttributesToClaimDisplay(claims: object, path: string[] = []):
...
+      compress: 1,
+    })
+
+    if (!savedImages.base64) {
+      return undefined
+    }
+
+    return `data:image/png;base64,${savedImages.base64}` as const
+  } catch (error) {
+    console.error('Error resizing image:', error)
```
berendsliedrecht, line 113: console.error?

packages/agent/src/openid4vc/registerDcApi.ts
```diff
@@ -64,16 +67,95 @@ function mapSdJwtAttributesToClaimDisplay(claims: object, path: string[] = []):
...
+    }
+    // In case of local image
+    else {
+      asset = ExpoAsset.Asset.fromModule(url)
+    }
+
+    return await resizeImageWithAspectRatio(asset)
+  } catch (error) {
+    // just ignore it, we don't want to cause issues with registering crednetials
+    console.error('Error resizing and retrieving cached image for DC API', {
```
berendsliedrecht, line 137: same here.

PR #345, opened 2025-06-04
feat: Relying Party Authentication

berendsliedrecht:


apps/easypid/src/app/(app)/_layout.tsx
```diff
@@ -172,7 +172,9 @@ export default function AppLayout() {
           <Stack.Screen name="activity/[id]" options={headerNormalOptions} />
           <Stack.Screen name="pinConfirmation" options={headerNormalOptions} />
           <Stack.Screen name="pinLocked" options={headerNormalOptions} />
-          <Stack.Screen name="federation" options={headerNormalOptions} />
+          <Stack.Screen name="trust/federation" options={headerNormalOptions} />
```
berendsliedrecht, line 175: They just use the same screen, but I split it up so it might be better in the future to have custom UI for this.

apps/easypid/src/features/wallet/FunkeFederationDetailScreen.tsx
```diff
@@ -1,92 +1,3 @@
...
-          </YStack>
-        </YStack>
-      </ScrollView>
-      <YStack btw="$0.5" borderColor="$grey-200" pt="$4" mx="$-4" px="$4" bg="$background">
-        <TextBackButton />
-      </YStack>
-    </FlexPage>
-  )
-}
+export const FunkeFederationTrustDetailScreen = FunkeTrustDetailScreen
```
TimoGlastra, line 3: i understand the separation, but it also seems like it's more complexity and in the end we just reuse the same page for all? So it could have just been 1 page with one component

packages/agent/src/invitation/handler.ts
```diff
@@ -32,15 +21,26 @@ import {
...
+} from '@credo-ts/openid4vc'
 import {
   getOfferedCredentials,
   getScopesFromCredentialConfigurationsSupported,
   preAuthorizedCodeGrantIdentifier,
 } from '@credo-ts/openid4vc'
-import { type Observable, filter, first, firstValueFrom, timeout } from 'rxjs'
-
+import { getOid4vcCallbacks } from '@credo-ts/openid4vc/build/shared/callbacks'
+import { getTrustedEntities } from '@easypid/utils/trustMechanism'
```
TimoGlastra, line 37: We shouldn't depend on easypid from the agent package  (and will lead to issues once we start the SDK work). 

I think need to provide the config from the constants to the calls to agent, that way we also keep the flexibility to dynamically change the trusted entities

(so i think we should move the `getTrustedEntities` to the agent, but keep the config/constants in the easypid and provide that to the `getTrustedEntities` as function param)

apps/easypid/src/utils/trustMechanism.ts
```diff
@@ -0,0 +1,235 @@
...
+        // trustedEntities.push({
+        //   entityId: trustedEntity.certificate,
+        //   organizationName: organizationName ?? trustedEntity.name,
+        // })
+
+        if (includeWalletAsTrustedEntity) {
+          trustedEntities.push({
+            organizationName: isParadymWallet() ? 'Paradym Wallet' : 'Funke Wallet',
+            entityId: '__',
+            logoUri: require('../../../../apps/easypid/assets/paradym/icon.png'),
```
TimoGlastra, line 216: We should probably make `logoUri` a param as well

PR #346, opened 2025-06-11
fix: remove navigation check to fix proximity flow

berendsliedrecht:


apps/easypid/app.config.js
```diff
@@ -54,7 +54,7 @@ APP_CONFIGS.FUNKE_WALLET.android.config = {
 }
 
 module.exports = () => {
-  const appType = process.env.EXPO_PUBLIC_APP_TYPE
+  const appType = process.env.EXPO_PUBLIC_APP_TYPE ?? 'FUNKE_WALLET'
```
TimoGlastra, line 57: Is the same default used in the code? Otherwise this could lead to weird behaviour between `prebuild` and `start`

apps/easypid/src/features/wallet/FunkeOfflineQrScreen.tsx
```diff
@@ -116,15 +115,10 @@ export function FunkeOfflineQrScreen() {
         deviceRequest: Buffer.from(data.deviceRequest).toString('base64'),
       })
     })
-
-    return () => {
-      if (!didNavigate) shutdownDataTransfer()
-    }
```
TimoGlastra, line 122: why don't we shutdown?

apps/easypid/app.config.js
TimoGlastra: The default in `appType.ts` is `PARADYM_WALLET`, so it's just a bit werid to have the default here be `FUNKE_WALLET`.

If we have a default, it should be consistent (and since we already use `PARADYM_WALLET` i think we should do that.

We should probably also figure out why the .env is not loaded when building with expo., I'd expect Expo to handle that

PR #348, opened 2025-06-18
feat(sdk): include agent initialization and providers

berendsliedrecht:
- Includes a way for the user to setup an agent
- Includes a context to add the agent to
- Includes a hook to retrieve the agent

# TODO

I would like to have the type of the agent be determined on wether `options.openid4vcconfiguration` and `options.didcommconfiguration` are provided. If they are, I would like the modules to be defined on the agent and otherwise be stripped. I have the types ready but I could not find a way to conditionally set them without wasting too much time. Ideas are welcome

# NOTE

All the providers are a direct copy, the initializeAgent is new-ish.

### DIFF
--- a/packages/sdk/package.json
+++ b/packages/sdk/package.json
@@ -22,16 +22,14 @@
 		"build": "tsup src/index.ts --format cjs,esm --dts --clean --sourcemap"
 	},
 	"dependencies": {
-		"@credo-ts/react-native": "*"
-	},
-	"devDependencies": {
-		"@openwallet-foundation/askar-react-native": "*"
+		"@credo-ts/askar": "catalog:",
+		"@credo-ts/core": "catalog:",
+		"@credo-ts/didcomm": "catalog:",
+		"@credo-ts/openid4vc": "catalog:",
+		"@credo-ts/react-native": "catalog:"
 	},
 	"peerDependencies": {
-		"@credo-ts/askar": "*",
-		"@credo-ts/core": "*",
-		"@credo-ts/didcomm": "*",
-		"@credo-ts/openid4vc": "*",
-		"react": "*"
+		"react": "catalog:",
+		"@openwallet-foundation/askar-react-native": "catalog:"
 	}
 }
--- a/packages/sdk/src/ParadymWalletSdk.tsx
+++ b/packages/sdk/src/ParadymWalletSdk.tsx
@@ -1,0 +1,82 @@
+import type { PropsWithChildren } from 'react'
+import { type FullAgent, type SetupAgentOptions, setupAgent } from './agent'
+import { ParadymWalletMustBeInitializedError } from './error'
+import { useCredentials } from './hooks/useCredentials'
+import { useCredentialsById } from './hooks/useCredentialsById'
+import { AgentProvider, useAgent } from './providers/AgentProvider'
+
+/**
+ *
+ * Options that will be used when initializing the Paradym Wallet SDK
+ *
+ * Make sure to instantiate the `ParadymWalletSdk` once and use the same instance within the application, via for example a `useContext`
+ *
+ */
+export type ParadymWalletSdkOptions = SetupAgentOptions
+
+export class ParadymWalletSdk {
+  private agent: FullAgent
+
+  public constructor(options: ParadymWalletSdkOptions) {
+    this.agent = setupAgent(options) as FullAgent
+  }
+
+  private assertAgentIsInitialized() {
+    if (!this.agent.isInitialized) throw new ParadymWalletMustBeInitializedError()
+  }
+
+  /**
+   *
+   * Initialized the wallet sdk and sets everything up for usage
+   *
+   * @note this must be called before any usage of the agent, such as retrieving credentials
+   *
+   */
+  public async initialize() {
+    await this.agent.initialize()
+  }
+
+  /**
+   *
+   * All available hooks provided by the wallet SDK
+   *
+   * @todo Ideally, we only want to return the hooks based on whether it is a didcomm/openid4vc agent
+   *       This can be done quite simple, but in order to provide correct types, it get very complex
+   *
+   */
+  public get hooks() {
+    this.assertAgentIsInitialized()
+
+    return {
+      useCredentials,
+      useCredentialsById,
+    }
+  }
+
+  /**
+   *
+   * Unstable hooks that can be used for more low-level functionality
+   *
+   * It is generally recommended to see if the desired output can be reached with `ParadymWalletSdk.hooks` first
+   *
+   */
+  public get internalHooks() {
+    this.assertAgentIsInitialized()
+
+    return {
+      useAgent,
+    }
+  }
+
+  /**
+   *
+   * Provider for the WalletSdk
+   *
+   * Wrap your application in this, if you want to leverage the provided `this.hooks`
+   *
+   */
+  public Provider({ children }: PropsWithChildren) {
+    this.assertAgentIsInitialized()
+    return <AgentProvider agent={this.agent}>{children}</AgentProvider>
+  }
+}
--- a/packages/sdk/src/agent.ts
+++ b/packages/sdk/src/agent.ts
@@ -1,12 +1,5 @@
 import { AskarModule } from '@credo-ts/askar'
-import {
-  Agent,
-  DidsModule,
-  KeyDerivationMethod,
-  type ModulesMap,
-  X509Module,
-  type X509ModuleConfigOptions,
-} from '@credo-ts/core'
+import { Agent, DidsModule, KeyDerivationMethod, X509Module, type X509ModuleConfigOptions } from '@credo-ts/core'
 import {
   ConnectionsModule,
   CredentialsModule,
@@ -22,17 +15,10 @@ import {
 import { OpenId4VcHolderModule } from '@credo-ts/openid4vc'
 import { agentDependencies } from '@credo-ts/react-native'
 import { askar } from '@openwallet-foundation/askar-react-native'
+import { ParadymWalletMustBeDidCommAgentError, ParadymWalletMustBeOpenId4VcAgentError } from './error'
 import { type LogLevel, logger } from './logger'
-import type { RequiredFields, StripOptionalAndUndefined } from './types'
 
-export type InitializeAgentOptions<AdditionalModules extends ModulesMap = ModulesMap> = {
-  /**
-   *
-   * Label used when using DIDComm to convey who you are
-   *
-   */
-  label: string
-
+export type SetupAgentOptions = {
   /**
    *
    * Unique identifier of your wallet storage
@@ -59,16 +45,6 @@ export type InitializeAgentOptions<AdditionalModules extends ModulesMap = Module
 
   /**
    *
-   * Custom agent modules that can be added
-   *
-   * @todo
-   * This is not yet supported
-   *
-   */
-  additionalAgentModules?: never
-
-  /**
-   *
    * Configuration for when OpenId4Vc is used
    *
    * @note by default, openid4vc is configured on the agent
@@ -87,7 +63,16 @@ export type InitializeAgentOptions<AdditionalModules extends ModulesMap = Module
    * @note to disable didcomm explicitly, pass in `false`
    *
    */
-  didcommConfiguration?: false
+  didcommConfiguration?:
+    | {
+        /**
+         *
+         * Label used for DIDComm connections to convey who you are to the other agent
+         *
+         */
+        label: string
+      }
+    | false
 }
 
 /**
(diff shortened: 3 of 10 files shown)


packages/sdk/package.json
```diff
@@ -21,5 +21,17 @@
...
 	},
-	"dependencies": {}
+	"dependencies": {
+		"@credo-ts/react-native": "*"
+	},
+	"devDependencies": {
+		"@openwallet-foundation/askar-react-native": "*"
+	},
+	"peerDependencies": {
+		"@credo-ts/askar": "*",
```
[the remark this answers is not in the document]
TimoGlastra: Yeah i think this shouldn't be peer dependencies.

And i think we can keep it consistent by using the PNPM catalog feature. So we can just add them as dependencies, and they will be correctly replaced during publishing.

packages/sdk/src/agent.ts
```diff
@@ -0,0 +1,171 @@
...
+   */
+  didcommConfiguration?: false
+}
+
+/**
+ *
+ * Function that sets up everything needed for the functionality required in a wallet.
+ *
+ */
+export const initializeAgent = async <AdditionalModules extends ModulesMap = ModulesMap>(
```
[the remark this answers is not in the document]
TimoGlastra: However I would make all the SDK commands available as non-hooks, and hook wrappers around it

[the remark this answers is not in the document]
TimoGlastra: Since not all wallets do everythign from wihtin a components, and you can't call a hook from outside of a component. We should not assume too much about the architecture

packages/sdk/package.json
```diff
@@ -21,5 +21,17 @@
 	"scripts": {
 		"build": "tsup src/index.ts --format cjs,esm --dts --clean --sourcemap"
 	},
-	"dependencies": {}
+	"dependencies": {
+		"@credo-ts/react-native": "*"
+	},
+	"devDependencies": {
+		"@openwallet-foundation/askar-react-native": "*"
```
TimoGlastra, line 28: This should be a peer, since it's a native dependency

packages/sdk/src/agent.ts
```diff
@@ -0,0 +1,171 @@
...
+  /**
+   *
+   * Configuration for when OpenId4Vc is used
+   *
+   * @note by default, openid4vc is configured on the agent
+   *
+   * @note to disable openid4vc, pass in `false`
+   *
+   */
+  openId4VcConfiguration?: ({} & X509ModuleConfigOptions) | false
```
TimoGlastra, line 79: Passing the X509module configuration to the openid4vc configuration does not make sense IMO.

I think we should stay away from Credo based configuration as much as possible in the public SDK. So e.g. additionalAgentModules should also not be exposed

packages/sdk/src/agent.ts
```diff
@@ -0,0 +1,171 @@
...
+   *
+   */
+  id: string
+
+  /**
+   *
+   * Key that will be used to securely open your wallet
+   *
+   * @security
+   * This property must dealt with in great care, such as receiving it from a biometrics check
```
TimoGlastra, line 48: Ideally we export utils for this from the SDK at some point as well. Basically what we have now in the secure unlock provider

packages/sdk/src/agent.ts
```diff
@@ -0,0 +1,171 @@
...
+import { type LogLevel, logger } from './logger'
+import type { RequiredFields, StripOptionalAndUndefined } from './types'
+
+export type InitializeAgentOptions<AdditionalModules extends ModulesMap = ModulesMap> = {
+  /**
+   *
+   * Label used when using DIDComm to convey who you are
+   *
+   */
+  label: string
```
TimoGlastra, line 34: This should be under DIDComm configuration

packages/sdk/src/agent.ts
berendsliedrecht: Yeah, thats the plan

packages/sdk/src/agent.ts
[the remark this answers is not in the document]
berendsliedrecht: Yeah makes sense, something like `getCredentials(agent, options)` and `useCredentials(options)`?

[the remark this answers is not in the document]
berendsliedrecht: Then again, not such a big difference than a useContext, wrt global variables.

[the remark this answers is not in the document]
TimoGlastra: Depends on whether we want to do functional or wrap the agent. 

For our API SDK we have:

```ts
const paradym = new Paradym({ apiKey: '' })
await paradym.openid4vc.offer()
```


I think that would be my approach for the SDK as well, but if we have good reason to keep it as separate methods it could work as well (just need to be extra careful in export then, since it's harder to detect what methods are availalbe)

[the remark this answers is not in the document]
TimoGlastra: It could be something like:

```ts
const walletSdk = new ParadymWalletSdk({

})

// now it's ready to sue
await walletSdk.initialize()
```

then if you also want to use the hooks, you can do something like this in a component:

```ts
return (
  <walletSdk.Provider>

  </walletSdk.Provider>
)
```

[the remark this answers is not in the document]
TimoGlastra: For hooks we could choose to use a separate scope, or directly on the SDK

```ts
const credentials = walletSdk.useCredentials()
```

berendsliedrecht:
> And change the config to:

But the modules need to be configured, how ever much we want to deviate from the credo API, we still need to pass things like `trustedCertificates` so a `extends boolean` would not work.

TimoGlastra:
This was simplified. We can do `extends MyVeryComplexObject | false = false` and then later do `DidcommEnabled extends false ? EmptyModulesMap : typeof didcommModules`.

packages/sdk/src/agent.ts
berendsliedrecht: I would love to have a trust::trustedCerts: boolean in the configuration, that is the goal. 

> And still, this does not explain why x509 config is part openid4vc-configuration?

Mainly combined this in there as I see two main agents `didcomm` or `openid` based. X.509 is only used for openid based agents so it made sense abstract it into the configuration there.

I fully agree that this exposes a too lowlevel API.

My three options, I think, are:

1. Leave it as-is
2. Do not let the user set it, and just have the callback be what we already have
3. Move already towards the ideal scenario where trust is defined by a simple boolean, enabling/disabling it.

Ofcourse 3 is the best, and that will be done, but preferably at a later stage. I am just really trying to focus with this on how it will be used and slightly less on all the exact configuration details.

berendsliedrecht:
Resolved the feedback I would like to address in this PR @TimoGlastra. It is mainly about this file: https://github.com/animo/paradym-wallet/blob/8522567dc599a99da52f829fda3b084c4ef84547/packages/sdk/src/ParadymWalletSdk.tsx

It introduces an extendable class which will only expose the simplified functionality, (hooks, provider, etc.).

The display information on the credentials and all will be done in a different PR (there are tickets for this), but this is more about checking in on the structure.

Usage would be something like

```typescript
import {ParadymWalletSdk} from '@paradym/wallet-sdk'


export const walletSdk = ParadymWalletSdk({walletOptions: {id: 'a', key: 'b'}})
await walletSdk.initialize()

// ..

import {walletSdk} from '..'

<walletSdk.Provider>
  <App />
</walletSdk.Provider

// .. 
import {walletSdk} from '..'

export const CredentialComponent = () => {
  const {credentials} = walletSdk.hooks.useCredentials()
}
```

packages/sdk/src/ParadymWalletSdk.tsx
```diff
@@ -0,0 +1,79 @@
...
+   *
+   * @todo Ideally, we only want to return the hooks based on whether it is a didcomm/openid4vc agent
+   *       This can be done quite simple, but in order to provide correct types, it get very complex
+   *
+   */
+  public get hooks() {
+    this.assertAgentIsInitialized()
+
+    return {
+      useAgent,
```
TimoGlastra, line 61: Maybe we can split these between `hooks` and `internalHooks`. And we make `internalHooks` not dependant on semver / are more advanced. 

We recommend users to use the `hooks`, and fallback to `internalHooks` with caution.

`useAgent` will be in `internalHooks`.

packages/sdk/src/ParadymWalletSdk.tsx
```diff
@@ -0,0 +1,79 @@
...
+  /**
+   *
+   * Options that setup the internal agent.
+   *
+   * You can define here whether you want to use `openid4vc`, `didcomm` or both
+   *
+   * @see `SetupAgentOptions` for the specific details
+   *
+   */
+  walletOptions: SetupAgentOptions
```
TimoGlastra, line 24: I think we need to start being consistent from the start, whether we use wallet or agent. Because the interface using `SetupAgentOptions`, but the property is called WalletOptions. 

Also, is there a specific reason to make these nested as wallet options? To me specifiying openid4vc top-level makes sense for a ParadymWalletSdk.

packages/sdk/src/agent.ts
```diff
@@ -0,0 +1,160 @@
...
+   *
+   */
+  key: string
+
+  /**
+   *
+   * The log level of the events that occur within the wallet
+   *
+   */
+  logLevel?: LogLevel
```
TimoGlastra, line 44: This would require us to export the Credo log-level. Do we want that? We could also just do `: 'debug'  | 'info'`, etc..

packages/sdk/src/agent.ts
```diff
@@ -0,0 +1,160 @@
...
+      logger: options.logLevel ? logger(options.logLevel) : undefined,
+    },
+    dependencies: agentDependencies,
+    modules,
+  })
+
+  if (didcommConfiguration) {
+    ;(agent as DidCommAgent).modules.didcomm.registerOutboundTransport(new HttpOutboundTransport())
+    ;(agent as DidCommAgent).modules.didcomm.registerOutboundTransport(new WsOutboundTransport())
+  }
```
TimoGlastra, line 114: I can't handle this syntax

```suggestion
  if (didcommConfiguration) {
    const didcommAgent  = agent as DidCommAgent
    didcommAgent.modules.didcomm.registerOutboundTransport(new HttpOutboundTransport())
    didcommAgent.modules.didcomm.registerOutboundTransport(new WsOutboundTransport())
  }
```

packages/sdk/src/hooks/useCredentials.tsx
```diff
@@ -0,0 +1,33 @@
...
+    const allRecords = [...w3cCredentialRecords, ...mdocRecords, ...sdJwtVcRecords]
+
+    // Sort by creation date
+    const sortedRecords = allRecords.sort((a, b) => b.createdAt.getTime() - a.createdAt.getTime())
+
+    return sortedRecords
+  }, [w3cCredentialRecords, sdJwtVcRecords, mdocRecords])
+
+  return {
+    credentials,
```
TimoGlastra, line 30: you mentioned adding display etc. Will we also abstract away the Credo records, and just return objects with the needed params?

packages/sdk/src/providers/W3cCredentialsProvider.tsx
TimoGlastra, line 1: Maybe we can also remove all the code that we copied from the agent package, so that Git sees it as a rewrite of the file name/path, and thus if we make a change to this file in main, it will try to merge it in this file.

I think this will make keeping it in sync with main a lot easier. No need to update anyhting in the wallet, just remove any code you copied in the agent package (also fine to do it in a separate PR, but if we do it for each PR, it's easier to keep track)

packages/sdk/src/agent.ts
berendsliedrecht: This is the exact code I had before, but it does require an allocation which I would like to avoid as it is just about the typing. I agree `;()` is quite ugly.

packages/sdk/src/providers/W3cCredentialsProvider.tsx
berendsliedrecht: Yeah, thats good. That means that we do not integrate the SDK yet, but just refer to the same internal functionality.

packages/sdk/src/hooks/useCredentials.tsx
berendsliedrecht: Yes, ideally I would like to return the bare minimum.

packages/sdk/src/ParadymWalletSdk.tsx
berendsliedrecht: Yeah I was thinking about this for too long. I just was not sure if there would be more options later that are not defined as `walletOptions`. I think having it top-level makes sense for now.

packages/sdk/src/agent.ts
TimoGlastra: Hmm.  I'm always of the opinion "clean code > optimized code" until performance becomes an issue, but fine by me

PR #350, opened 2025-06-23
Wallet sdk credential and proof plus display

berendsliedrecht:
## Credential and proof display

I have migrated quite some functionality from `@packages/agent` to `@paradym/wallet-sdk`. This mainly includes the display functionality for credentials and proofs. It also includes a lot of hooks, invitation functionality and utils. The `easypid` app now uses some of this new functionality from the wallet sdk, but it has to be refactored to use the `ParadymWalletSdk` instance, so the usage is not too relevant. I mainly changed it so that migrating to usage of the `ParadymWalletSdk` is easier, as all the types already line up. 

I also used `Error` internally, as all cases that were reached were fatal. At a higher level, `ParadymWalletSdk`, we can wrap this and convert into an object with a `success` or something equivalent.

Git also did not correctly remap some files, even when using `git mv` sadly.

There are still some tasks related to this, like improving the DX of the proof display attributes, but I would like to do this in a new PR as +3.3k and -2.6k is a bit much.


apps/easypid/src/agent/index.ts
```diff
@@ -4,9 +4,9 @@ import {
   type ParadymAppAgent,
   isEasyPIDAgent,
   isParadymAgent,
-  useAgent,
 } from '@package/agent'
 import { useSecureUnlock as _useSecureUnlock } from '@package/secure-store/secureUnlock'
+import { useAgent } from '@paradym/wallet-sdk/src/providers/AgentProvider'
```
TimoGlastra, line 9: I updated all the paths in the expo update pr to not use `/src` anymore. By adding a tsconfig and metro/babel mapping.

I think for consistency would be great to do it here as well. (also, but mabye for future, it might be nice to look at export mapping in package json so we can have multiple export for different features, e.g. `@paradym/wallet-sdk/providers` , etc..)

apps/easypid/src/features/activity/activityRecord.ts
```diff
@@ -1,17 +1,18 @@
...
   useWalletJsonRecord,
 } from '@package/agent'
-
+import type { BaseAgent } from '@paradym/wallet-sdk/src/agent'
+import {
+  getDisclosedAttributeNamesForDisplay,
+  getUnsatisfiedAttributePathsForDisplay,
+} from '@paradym/wallet-sdk/src/display/common'
+import type { DisplayImage } from '@paradym/wallet-sdk/src/display/credential'
+import type { FormattedSubmission } from '@paradym/wallet-sdk/src/format/submission'
```
TimoGlastra, line 15: What is the idea for users of the SDK? Will they import just from `@paradym/wallet-sdk` and are these internal utils we use?

apps/easypid/src/features/didcomm/DidCommNotificationScreen.tsx
```diff
@@ -54,8 +55,8 @@ export function DidCommNotificationScreen() {
   const onComplete = () => handleNavigation('replace')
 
   const onConnectionAccept = async () => {
-    const result = await acceptConnection()
-    if (result.success) {
```
TimoGlastra, line 58: I'm kinda a fan of the `success` and then having data or error, since:
- try/catch is a bit funky in react
- it allows you to provide more specific errors metadata (like for parse invitation what the error type is). If we want to do that with error you need more documentation on the error types, while with typescript the return type can already include this. 
- You need to either check if error is instance of error, or cast as an error. Bot are suboptimal

So if we are changing this model we should also think about how the user of the SDK will deal with errors.

apps/easypid/src/use-cases/ReceivePidUseCaseCFlow.ts
```diff
@@ -1,16 +1,17 @@
-import type { MdocRecord } from '@credo-ts/core'
+import type { MdocRecord, SdJwtVcRecord } from '@credo-ts/core'
 import { pidSchemes } from '@easypid/constants'
 import {
   BiometricAuthenticationError,
   OpenId4VciAuthorizationFlow,
-  type SdJwtVcRecord,
   receiveCredentialFromOpenId4VciOffer,
   resolveOpenId4VciOffer,
+  storeCredential,
```
TimoGlastra, line 8: shouldn't storeCredential be handled by the SDK as well, or is that a follow up task?

apps/easypid/src/features/activity/activityRecord.ts
berendsliedrecht: Same as the comment above.

Types might always be exported from `src/index.ts`, but I need to find the best pattern for that.

packages/sdk/src/hooks/useMessagePickup.tsx
```diff
@@ -0,0 +1,33 @@
...
+import type { DidCommAgent } from '../agent'
+import { initiateMessagePickup, stopMessagePickup } from '../didcomm/mediation'
+
+/**
+ * Hook to enable message pickup from the mediator.
+ *
+ * You can use the `isEnabled` config property to enable/disable message pickup.
+ * This is useful if e.g. there's no internet connection, or mediation has not been setup yet
+ */
+export function useMessagePickup({
```
TimoGlastra, line 11: Might be good to call this `useDidcommMessagePickup`

packages/sdk/src/hooks/useCredentials.tsx
```diff
@@ -1,30 +1,70 @@
...
- * @todo add Display information to the records
- *
- */
 export const useCredentials = ({
   removeCanonicalRecords = true,
+  credentialCategory,
 }: {
   removeCanonicalRecords?: boolean
+  // TODO: remove this?
+  credentialCategory?: CredentialCategoryMetadata['credentialCategory']
```
TimoGlastra, line 14: Why remove it? I think we're using this to retrieve the PID record

packages/sdk/src/hooks/useCredentialRecordById.tsx
```diff
@@ -0,0 +1,10 @@
+import { useCredentialRecords } from './useCredentialRecords'
+
+export const useCredentialRecordById = (id: string) => {
```
TimoGlastra, line 3: I think for all these hooks we need to update it to use the for display variant by default (which also includes the record i think?)

packages/sdk/src/display/mapAttributes.ts
```diff
@@ -0,0 +1,35 @@
...
+
+type MappedAttributesReturnType =
+  | string
+  | number
+  | boolean
+  | { [key: string]: MappedAttributesReturnType }
+  | null
+  | undefined
+  | Array<MappedAttributesReturnType>
+export function recursivelyMapAttributes(value: unknown): MappedAttributesReturnType {
```
TimoGlastra, line 13: I think this is only used for mdoc right? If so we should probably make it clear in the function name

packages/app/package.json
```diff
@@ -8,17 +8,20 @@
 		"@package/scanner": "workspace:*",
 		"@package/ui": "workspace:*",
 		"@package/utils": "workspace:*",
+		"@paradym/wallet-sdk": "workspace:*",
 		"@react-native-community/netinfo": "catalog:",
 		"@tamagui/animations-react-native": "catalog:",
 		"@tamagui/colors": "catalog:",
 		"@tamagui/font-inter": "catalog:",
 		"@tamagui/lucide-icons": "catalog:",
+		"@tamagui/portal": "catalog:",
```
TimoGlastra, line 17: I think i removed `@tamagui/portal` in a recent PR. Do we import this anywhere?

packages/app/package.json
```diff
@@ -8,17 +8,20 @@
...
 		"@tamagui/shorthands": "catalog:",
 		"@tamagui/themes": "catalog:",
 		"@tanstack/react-query": "catalog:",
 		"expo-constants": "catalog:",
 		"expo-haptics": "catalog:",
 		"expo-linking": "catalog:",
+		"expo-modules-autolinking": "catalog:",
```
TimoGlastra, line 24: Do we need this? I think this is already installed by expo (i removed it in a recent PR)

TimoGlastra (approved):

packages/sdk/src/hooks/useCredentials.tsx
berendsliedrecht: Yeah mainly because we only really use it for the PID. Just wondering about how relevant that is in a general SDK.

packages/sdk/src/hooks/useCredentialRecordById.tsx
berendsliedrecht: the `useCredentials` now returns display information, and the `useCredentialRecords` only the record. `useCredentials` will be a public hook and `useCredentialRecords` will be an internal hook

PR #358, opened 2025-07-18
feat: usage of wallet sdk in easypid app

berendsliedrecht:
- **chore: moved all hooks and seperated the refresh pid method**
- **chore: paradymwalletsdk provider and using in the easypid app**
- **chore: moved secure unlock to the paradym wallet sdk**
- **chore: use return context instead of errors in the sdk**

### DIFF
--- a/apps/easypid/src/sdk/paradymWalletSdk.tsx
+++ b/apps/easypid/src/sdk/paradymWalletSdk.tsx
@@ -13,12 +13,12 @@ export function getWalletId(walletKeyVersion: number) {
 
 export const paradymWalletSdk = () => {
   if (!paradym) {
-    throw new Error('Paradym Wallet Sdk is not yet created. Call `initializeParadymWalletSdk()` first')
+    throw new Error('Paradym Wallet SDK is not yet created. Call `initializeParadymWalletSdk()` first')
   }
 
   if (!paradym.agent.isInitialized) {
     throw new Error(
-      'Paradym Wallet Sdk is not yet initialized. Something went wrong while calling `initializeParadymWalletSdk()`'
+      'Paradym Wallet SDK is not yet initialized. Something went wrong while calling `initializeParadymWalletSdk()`'
     )
   }
 


packages/sdk/src/hooks/useActivities.tsx
```diff
@@ -0,0 +1,20 @@
...
+import { type ActivityRecord, activityStorage } from '../storage/activities'
+
+export const useActivities = ({ filters }: { filters?: { entityId?: string } } = {}) => {
+  const { record, isLoading } = useWalletJsonRecord<ActivityRecord>(activityStorage.recordId)
+
+  const activities = useMemo(() => {
+    if (!record?.activities) return []
+
+    return [...record.activities]
+      .sort((a, b) => new Date(b.date).getTime() - new Date(a.date).getTime())
```
hacdias, line 12: Maybe a nitpick, but do we need to do this transformation (`new Date(x).getTime()`) every time the sorting function is called? Is it not possible to use `.date` directly?

packages/sdk/src/invitation/shareProof.ts
```diff
@@ -115,8 +124,8 @@ export const shareProof = async ({
 
     return result
   } catch (error) {
-    // Handle biometric authentication errors
-    throw BiometricAuthenticationError.tryParseFromError(error) ?? error
+    // TODO
+    throw Error('TODO')
```
hacdias, line 128: Missing error?

packages/sdk/src/openid4vc/getCredentialsForProofRequest.ts
```diff
@@ -70,5 +80,16 @@ export const getCredentialsForProofRequest = async ({
     authorizationRequest: resolved.authorizationRequestPayload,
     formattedSubmission,
     transactionData: resolved.transactionData,
+    // TODO: add verifier with trust integration
+    trustMechanism: undefined,
+    verifier: { entityId: 'TODO', hostName: 'TODO', logo: { url: 'TODO' }, name: 'TODO' } as {
```
hacdias, line 85: Missing?

hacdias (commented):
I see this PR is to be merged onto the `wallet-sdk` branch, and I noticed there's quite a few `TODO(sdk)` (I did not highlight all).These will come in a different PR, right?

There's also some other smaller things I noticed

apps/easypid/src/app/onboarding/index.tsx
```diff
@@ -39,8 +41,9 @@ export default function OnboardingScreens() {
 
     setHasResetWallet(true)
     router.setParams({ reset: 'false' })
-    resetWallet(secureUnlock)
-  }, [reset, secureUnlock, hasResetWallet])
+    // TODO(sdk): move to sdk
```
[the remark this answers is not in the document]
berendsliedrecht: Yes :)

apps/easypid/src/features/share/FunkeOpenIdPresentationNotificationScreen.tsx
```diff
@@ -50,7 +49,8 @@ export function FunkeOpenIdPresentationNotificationScreen() {
     getCredentialsForProofRequest({
       agent,
       uri: params.uri,
-      trustedX509Entities,
+      // TODO: add back when trust
+      // trustedX509Entities,
```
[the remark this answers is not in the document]
berendsliedrecht: Yes!

packages/app/src/features/notifications/DidCommCredentialNotificationScreen.tsx
```diff
@@ -27,7 +22,7 @@ export function DidCommCredentialNotificationScreen({
   }
 
   const onCredentialAccept = async () => {
-    await acceptCredential()
+    void acceptCredential()
```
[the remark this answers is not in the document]
berendsliedrecht: For the accept, we handle everything in the `then` `catch` and `finally`. We mark it as `void` here as we do not want to await the code for the next line, as we handle everything in `then`. We don't need to mark it as `void`, but I prefer to do that to show it is a promise, without `awaiting` it.

And the decline, the output does not matter, i.e. we do not need to then/catch, and we use `void` so the app does not hang and we continue declining in the background.

packages/sdk/src/hooks/useActivities.tsx
berendsliedrecht: It might be possible, but we store the date internally in the agent and as react state. This means we have to serialize it (currently to an ISO string) and deserialize it.

This means that when we store then we have to call `toIsoString()` and when we fetch it we call `new Date()`. This would slightly improve the sort call, as we do not have to recreate the date for all instances, but the performance gain seems minimal. 

It would be a good cleanup/optimalization, but I would give it a very low prio.

packages/sdk/src/openid4vc/getCredentialsForProofRequest.ts
berendsliedrecht: trust + verifier info will come in different PR.

PR #362, opened 2025-08-04
fix: redirect works when app is longer in background

berendsliedrecht:
replaces #337 

- Request authentication when a deeplink is being used (same fix as in #337)
- Fixes the issue when the app is longer in the background that the deeplink still works, comes with the side effect that when the app is not so long in the background that it needs to re-auth, which I think is okay. Mainly, because re-authenticating within an such a fraud sensitive platform is not a bad thing. Most people, if not all, use biometrics and make the authentication extremely fast.

I could not test it with the Ausweis flow as I do not have a card and I could not find the website anymore. I think the deeplink working has a priority over the ausweis flow for now.

### DIFF
--- a/apps/easypid/src/app/+native-intent.tsx
+++ b/apps/easypid/src/app/+native-intent.tsx
@@ -1,5 +1,6 @@
 import 'fast-text-encoding'
 
+import { TypedArrayEncoder } from '@credo-ts/core'
 import { appScheme } from '@easypid/constants'
 import { parseInvitationUrl } from '@package/agent'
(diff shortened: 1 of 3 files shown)


berendsliedrecht:
> @berendsliedrecht quick q: are the other changes from #337 not needed?

apps/easypid/src/features/menu/FunkeMenuScreen.tsx seems to contain a new menu entry to lock the app, which does not seem relevant for deeplinking

I did add the other fix now as well. I think it had to do with navigating back when a deeplink is used, i.e. no previous route, and if that is the case we just navigate to `/`.

PR #367, opened 2025-08-14
feat: added trust in the sdk

berendsliedrecht:
depends on #364 

- Added trust back with configuration
- Works just as before

configuration: 

```typescript
export const paradymWalletSdkOptions: SetupParadymWalletSdkOptions = {
  logLevel: LogLevel.debug,
  openId4VcConfiguration: {
    // TODO: maybe we can remove this because of the trustMechanisms?
    trustedCertificates: trustedX509Certificates as [string, ...string[]], 
  },
  trustMechanisms: [
    { trustMechanism: 'eudi_rp_authentication', trustList: eudiTrustList, trustedX509Entities },
    { trustMechanism: 'openid_federation', trustedEntityIds },
    { trustMechanism: 'x509', trustedX509Entities },
    { trustMechanism: 'did' },
  ],
}
```

big question, what to do when a trust mechanism is used, but not configured? Error or something else? Maybe stop continue flow, but add a return value to a context object of `isSupportedTrustMechanismUsed` or something equivalent?


apps/easypid/src/hooks/useShouldUsePinForPresentation.tsx
```diff
@@ -1,4 +1,4 @@
-import type { CredentialsForProofRequest } from '@package/agent'
+import type { CredentialsForProofRequest } from '@paradym/wallet-sdk/openid4vc/getCredentialsForProofRequest'
```
TimoGlastra, line 1: should we export from root?

packages/sdk/src/ParadymWalletSdk.tsx
```diff
@@ -22,32 +23,82 @@ import { type InvitationResult, parseDidCommInvitation, parseInvitationUrl } fro
...
   }
 
   private assertAgentIsInitialized() {
     if (!this.agent.isInitialized) throw new ParadymWalletMustBeInitializedError()
   }
 
+  public get walletId() {
+    const walletKeyVersion = secureWalletKey.getWalletKeyVersion()
+
+    // TODO(sdk): how do we want to deal with the `easypid` wallet id?
```
TimoGlastra, line 64: I think for Paradym we also have a different prefix? OR is it also easypid-wallet? I would pick a different default and make it configurable (so we can provide a custom one for easypid)

packages/sdk/src/ParadymWalletSdk.tsx
```diff
@@ -22,32 +23,82 @@ import { type InvitationResult, parseDidCommInvitation, parseInvitationUrl } fro
...
+    return `easypid-wallet-${walletKeyVersion}`
+  }
+
+  public async reset(additionalResetCb?: () => Promise<void> | void) {
```
TimoGlastra, line 68: I'm never too sure what a callback that is called at the end adds. 

It's easier IMO to just do:
```
sdk.reset()
myExtraResetSteps()
```

apps/easypid/src/hooks/useShouldUsePinForPresentation.tsx
berendsliedrecht: Yeah I want to fix the exports all in the end. Bit difficult to see what to export now and what not to, as some things still have to be made simpler.

packages/sdk/src/ParadymWalletSdk.tsx
berendsliedrecht: Configurable with default is 👍

packages/sdk/src/ParadymWalletSdk.tsx
berendsliedrecht: Yeah it was left over from when something happened afterwards, but can be removed again.

apps/easypid/src/features/share/FunkeMdocOfflineSharingScreen.tsx
```diff
@@ -24,7 +24,7 @@ export function FunkeMdocOfflineSharingScreen({
   sessionTranscript,
   deviceRequest,
 }: FunkeMdocOfflineSharingScreenProps) {
-  const { agent } = useOpenId4VcAgent()
+  const { paradym } = useParadym('unlocked')
```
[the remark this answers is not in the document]
berendsliedrecht: Wait that does not solve the issue.

PR #373, opened 2025-08-25
feat(sdk): receive openid4vci credentials

berendsliedrecht:
Converted most of the complexity of the FunkeOpenIdCredentailNotification page into some `paradym.opendi4vc` calls. Any ideas to make it even easier are very much welcome. 

Do we like the pattern of the didcomm hooks? `const {acceptCred, retrieveCredsForProof, etc.} = useOpenId4VciActions()`

Is there something else we can abstract away? I still have to migrate some other vci functionality, but I wanted to get this correct, and resolve all feedback, and then move to the other functions. So basically, the pr is only about FunkeOpenIdCredentialNotification for API usage and code quality everywhere else.


apps/easypid/src/features/onboarding/onboardingContext.tsx
```diff
@@ -505,7 +505,7 @@ export function OnboardingContextProvider({
           )
 
           const { display } = getCredentialForDisplay(credential)
-          await addReceivedActivity(paradym.paradym.agent, {
+          await addReceivedActivity(paradym.paradym, {
```
berendsliedrecht, line 508: These calls should also be in the sdk at some point.

apps/easypid/src/features/receive/FunkeOpenIdCredentialNotificationScreen.tsx
```diff
@@ -109,188 +62,40 @@ export function FunkeCredentialNotificationScreen() {
     [isDevelopmentModeEnabled]
   )
 
-  const shouldUsePinForPresentation = useShouldUsePinForSubmission(credentialsForRequest?.formattedSubmission)
-  const preAuthGrant =
-    resolvedCredentialOffer?.credentialOfferPayload.grants?.['urn:ietf:params:oauth:grant-type:pre-authorized_code']
-  const txCode = preAuthGrant?.tx_code
+  const shouldUsePinForPresentation = useShouldUsePinForSubmission(
+    resolvedOffer?.flow === 'auth-presentation-during-issuance'
```
berendsliedrecht, line 66: Seperated auth and auth with presentation as the UI flow is quite different.

apps/easypid/src/features/receive/FunkeOpenIdCredentialNotificationScreen.tsx
```diff
@@ -109,188 +62,40 @@ export function FunkeCredentialNotificationScreen() {
...
-        setResolvedAuthorizationRequest(resolvedAuthorizationRequest)
+    if (resolvedOffer) return
+
+    paradym.openid4vc
+      .resolveCredentialOffer({
+        offer: params,
+        authorization,
+      })
+      .then((result) => {
+        setResolvedOffer(result)
```
berendsliedrecht, line 80: I could also have the `resolveOffer` object contain a function which does the `acquireCredential`. Might be a bit overkill, but we think that that makes it easier, I can change it instead of a paradym sdk call.

apps/easypid/src/features/receive/FunkeOpenIdCredentialNotificationScreen.tsx
```diff
@@ -463,19 +334,19 @@ export function FunkeCredentialNotificationScreen() {
...
-        isAuthFlow && shouldUsePinForPresentation
+        resolvedOffer.flow === 'auth-presentation-during-issuance' && shouldUsePinForPresentation
           ? {
               step: 'pin-enter',
               progress: 82.5,
               screen: <PinSlide key="pin-enter" isLoading={isSharingPresentation} onPinSubmit={onPresentationAccept} />,
             }
           : undefined,
-        isPreAuthWithTxFlow
+        resolvedOffer.flow === 'pre-auth' && resolvedOffer.txCodeInfo
```
berendsliedrecht, line 344: Could also add a flow of `pre-auth-with-transaction` as the UI is also quite different there.

packages/sdk/src/ParadymWalletSdk.tsx
```diff
@@ -260,4 +278,25 @@ export class ParadymWalletSdk {
       return { success: false, message: error instanceof Error ? error.message : `${error}` }
     }
   }
+
+  public get openid4vc() {
+    return {
+      resolveCredentialOffer: (options: Omit<ResolveCredentialOfferOptions, 'paradym'>) =>
+        resolveCredentialOffer({ ...options, paradym: this }),
+
+      // TODO(sdk): do we just want a acquireCredentials, and handle it according to the input?
```
hacdias, line 287: I like that idea, since it'd abstract more to the user.

PR #383, opened 2025-09-29
feat: support for w3c v2 and async issuance

hacdias:
This PR adds support for:

* W3C V2 Credentials: this hasn't been tested, since we don't support this format in our Playground. Assuming the work in Credo works correctly, the wallet implementation should also work correctly. We need to test it.
* Async Issuance: this has been tested against the `main` version of the Playground.
* Translations still need to be added, but I'll do the translation with AI **once** we have the final English copy.

Things still to do/fix:

- [x] mDOC presentation
- [x] mDOC issuance
- [x] Added credential text doesn't make it clear that it is still pending
- [x] Make sure text is not truncated with translations on the "Deferred credential notification" page
- [x] Make called 'deferred' pending in all screens and update translations

Some screenshots related to async issuance:

| Adding deferred credential | Notification page | Notification page with failure |
| - | - | - | 
| IMG_5046 | IMG_5049|  IMG_5050 | 
| Deleting deferred credential  | Activity list | Notifications list |
| IMG_5051 |  IMG_5052 | IMG_5048  |
| Adding credential success  | - | - |
| IMG_5047  | |

### DIFF
--- a/apps/easypid/src/features/receive/DeferredCredentialNotificationScreen.tsx
+++ b/apps/easypid/src/features/receive/DeferredCredentialNotificationScreen.tsx
@@ -20,35 +20,36 @@ import {
   Heading,
   HeroIcons,
   InfoButton,
+  Loader,
   Paragraph,
   ScrollView,
   Stack,
   YStack,
 } from '@package/ui'
 import { useLocalSearchParams, useRouter } from 'expo-router'
-import { useState } from 'react'
+import { useEffect, useState } from 'react'
 import { useSafeAreaInsets } from 'react-native-safe-area-context'
 
 type Query = { deferredCredentialId: string }
 
 export function DeferredCredentialNotificationScreen() {
   const router = useRouter()
-  const { withHaptics } = useHaptics()
-  const { bottom } = useSafeAreaInsets()
-  const { handleScroll, isScrolledByOffset, scrollEventThrottle } = useScrollViewPosition()
-  const { i18n, t } = useLingui()
   const [isSheetOpen, setIsSheetOpen] = useState(false)
+  const { i18n, t } = useLingui()
+  const { handleScroll, isScrolledByOffset, scrollEventThrottle } = useScrollViewPosition()
+  const { bottom } = useSafeAreaInsets()
+  const { withHaptics } = useHaptics()
+
   const { deferredCredentialId } = useLocalSearchParams<Query>()
-  const { deferredCredentials } = useDeferredCredentials()
-
+  const { deferredCredentials, isLoading } = useDeferredCredentials()
   const deferredCredential = deferredCredentials?.find((dc) => dc.id === deferredCredentialId)
-  if (!deferredCredential) {
-    router.back()
-    return
-  }
+  const nextCheck = deferredCredential ? getDeferredCredentialNextCheckAt(deferredCredential) : undefined
 
-  const { response, issuerMetadata } = deferredCredential
-  const nextCheck = getDeferredCredentialNextCheckAt(deferredCredential)
+  useEffect(() => {
+    if (!isLoading && !deferredCredential) {
+      router.back()
+    }
+  }, [router, isLoading, deferredCredential])
 
   useHeaderRightAction({
     icon: <HeroIcons.Trash />,
@@ -56,16 +57,18 @@ export function DeferredCredentialNotificationScreen() {
     renderCondition: true,
   })
 
-  const credentialDisplay = getCredentialDisplayWithDefaults(
-    getOpenId4VcCredentialDisplay(
-      extractOpenId4VcCredentialMetadata(response.credentialConfiguration, {
-        display: issuerMetadata.credentialIssuer?.display,
-        id: issuerMetadata.credentialIssuer?.credential_issuer,
-      })
-    )
-  )
+  const credentialDisplay = deferredCredential
+    ? getCredentialDisplayWithDefaults(
+        getOpenId4VcCredentialDisplay(
+          extractOpenId4VcCredentialMetadata(deferredCredential.response.credentialConfiguration, {
+            display: deferredCredential.issuerMetadata.credentialIssuer?.display,
+            id: deferredCredential.issuerMetadata.credentialIssuer?.credential_issuer,
+          })
+        )
+      )
+    : undefined
 
-  return (
+  return deferredCredential && credentialDisplay ? (
     <>
       <FlexPage p={0} gap={0}>
         <YStack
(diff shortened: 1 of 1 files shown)


apps/easypid/.env.development
```diff
@@ -1,2 +1,2 @@
 EXPO_PUBLIC_WALLET_SERVICE_PROVIDER_URL=
-EXPO_PUBLIC_APP_TYPE=FUNKE_WALLET
\ No newline at end of file
+EXPO_PUBLIC_APP_TYPE=PARADYM_WALLET
```
TimoGlastra, line 2: I think we have FUNKE_WALLET as default in main right?

packages/translations/src/commonMessages.ts
```diff
@@ -278,6 +293,32 @@ export const commonMessages = {
...
+    message: 'Delete deferred credential?',
+    comment: 'Title for confirmation dialog to delete a deferred credential',
+  }),
+  deleteDeferredCredentialDescription: (name: string) =>
+    defineMessage({
+      id: 'common.deleteDeferredCredentialConfirmDescription',
+      message: `By deleting ${name}, you will no longer be able to retrieve your credential from the issuer.`,
+      comment: 'Description in confirmation dialog explaining deferred credential deletion',
+    }),
+  deleteDeferredCredentialConfirm: defineMessage({
```
TimoGlastra, line 307: I think this is still using the old ID?

packages/translations/src/commonMessages.ts
```diff
@@ -227,6 +232,11 @@ export const commonMessages = {
     message: 'Document signed',
     comment: 'Shown when signing a document succeeded',
   }),
+  signingPending: defineMessage({
```
TimoGlastra, line 235: Signing is during presentation, which doesn't support deferred issuance, so not sure when this text is used?

packages/agent/src/storage/deferredCredentialStore.ts
```diff
@@ -0,0 +1,120 @@
...
+    if (!record) {
+      throw new Error('No deferred credential record found')
+    }
+
+    record.deferredCredentials = record.deferredCredentials.filter((d) => d.id !== id)
+    await _deferredCredentialStorage.update(agent, record)
+  },
+}
+
+export const useDeferredCredentials = () => {
```
TimoGlastra, line 79: I agree we should make this a record per item at some point. The generic records would work fine for this.Not sure why we didn't do that immediatley tbh. Probably because this was quicker 🤷

packages/agent/src/openid4vc/deferredCredentialRecord.ts
```diff
@@ -0,0 +1,119 @@
...
+      })
+
+      await deleteDeferredCredential(agent, deferredCredential.id)
+    } else {
+      await updateDeferredCredential(agent, {
+        id: deferredCredential.id,
+        lastCheckedAt: new Date().toISOString(),
+        // Update access token with a new one if we got a new refresh token.
+        // Otherwise, keep the existing one.
+        accessToken: accessToken.refreshToken
```
TimoGlastra, line 94: The access token response returns `expires_in`. Should we somehow store this on the record to know:
- when the deferred issuance failed due to the token having expired (in case no refresh token)
- whether we should retrieve a new access token or use the existing one (in case there's a refresh token(

packages/agent/src/invitation/handler.ts
```diff
@@ -710,6 +746,7 @@ export async function acceptOutOfBandInvitation<FlowType extends 'issue' | 'veri
   try {
     const receiveInvitationResult = await agent.modules.outOfBand.receiveInvitation(invitation, {
       reuseConnection: true,
+      label: isParadymWallet() ? 'Paradym Wallet' : 'EasyPID',
```
TimoGlastra, line 749: same here

apps/easypid/src/features/receive/DeferredCredentialNotificationScreen.tsx
```diff
@@ -0,0 +1,153 @@
...
+                    comment: 'Title of the error box explaining deferred credentials',
+                  })}
+                  description={t({
+                    id: 'deferredCredentials.failedToRetrieveCardDescription',
+                    message: 'An error occurred while retrieving the deferred card from the issuer.',
+                    comment: 'Description of the error box explaining deferred credentials',
+                  })}
+                />
+              )}
+              <InfoButton
```
TimoGlastra, line 122: I think it would be good to share the `interval` with the user. E.g. `The issuer of the card indicated you should check back in 2 hours`

apps/easypid/src/features/receive/FunkeOpenIdCredentialNotificationScreen.tsx
```diff
@@ -141,7 +149,8 @@ export function FunkeCredentialNotificationScreen() {
       configurationId: string,
       resolvedAuthorizationRequest?: OpenId4VciResolvedAuthorizationRequest
     ) => {
-      const credentialResponses = await receiveCredentialFromOpenId4VciOffer({
+      console.log('ClientId', resolvedAuthorizationRequest ? authorization.clientId : undefined)
```
TimoGlastra, line 152: should use agent logger

apps/easypid/src/features/receive/slides/CredentialRetrievalSlide.tsx
```diff
@@ -212,7 +218,16 @@ export const CredentialRetrievalSlide = ({
               >
                 {scrollViewHeight && isAllowedToAccept ? (
                   <AnimatedStack key="credential-attributes" entering={FadeIn.duration(200)}>
-                    <CredentialAttributes attributes={attributes} />
+                    {attributes ? (
+                      <CredentialAttributes attributes={attributes} />
+                    ) : (
+                      <Paragraph>
+                        <Trans id="receiveCredential.deferredCredentialParagraph">
+                          Your credentials are not yet ready. We will check once in a while in the background and notify
```
TimoGlastra, line 226: Same here, i think we should integrate the interval here, so the user has a sense of when to expect the credential.

hacdias: I think integrating the interval here is a bit more complex than in the other screen. Especially considering we have to deal with three variations of this message (no interval, interval which falls within today, interval which falls outside of today). I think having it in the notification page  should be enough for now.

packages/agent/src/openid4vc/deferredCredentialRecord.ts
hacdias: I updated the fetching logic to only get a refresh token if the access token is indeed expired.

hacdias:
> did you use thte AI instruction to generate the translations?

Yes

hacdias:
I'm merging this one into `feat/latest-credo`. We can see how to go from there.

PR #394, opened 2025-11-05
fix(android): make it work in production again

berendsliedrecht:

### DIFF
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -3,7 +3,7 @@ packages:
   - 'apps/*'
 
 catalog:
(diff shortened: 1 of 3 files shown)


pnpm-workspace.yaml
```diff
@@ -3,7 +3,7 @@ packages:
   - 'apps/*'
 
 catalog:
-  expo: ~53.0.15
+  expo: 53.0.13
```
berendsliedrecht: @TimoGlastra if you want, I can test a bit more versions, but it depends a bit when we want to go to 54.

PR #438, opened 2025-12-08
feat: update Credo to 0.6, Expo 54

TimoGlastra:
WIP, needs testing.

Updates Credo to 0.6, removes federation for now, until we have properly integrated it into Credo

It depends on 0.6.1 alpha as we need some imports that are not imported in 0.6. I think we can very soon make a 0.6.1 release.

The batch issuance is also upated to Credo's logic, this is definitely better, but we do need to refactor it a bit to work better (refresh should be run continuously)

Also adds an option to bypass biometrics setup, and allows to enable/disable it in settings. (this should probably have been a separate PR)

Todo:
- [x] Make Argon parameters configurable in Askar (https://github.com/openwallet-foundation/askar/issues/446)
- [x] Updrade Askar JS wrapper and integrate new release
- [x] Fix the "WebAssembly.Module(): BufferSource argument is empty" error
- [x] Fix no internet connectin popup always showing on load
- [x] Error with key id (it tries to present with legacy key id)
- [x] Fix incorrect PIN entered (wrong askar store key) new error thrown in 0.6
- [ ] test that a wallet using the previous version can still derive the wallet key correctly based on the new argon implementation

### DIFF
--- a/apps/easypid/src/crypto/WalletServiceProviderClient.ts
+++ b/apps/easypid/src/crypto/WalletServiceProviderClient.ts
@@ -13,6 +13,7 @@ import { agentDependencies } from '@credo-ts/react-native'
 import { askar } from '@openwallet-foundation/askar-react-native'
 import type { EasyPIDAppAgent } from '@package/agent'
 import { secureWalletKey } from '@package/secure-store/secureUnlock'
+import { getWalletId } from '../agent/initialize'
 import { InvalidPinError } from './error'
 import { deriveKeypairFromPin } from './pin'
 
@@ -23,7 +24,8 @@ export const setWalletServiceProviderPin = async (pin: Array<number>, validatePi
   if (validatePin) {
     const walletKeyVersion = secureWalletKey.getWalletKeyVersion()
     const walletKey = await secureWalletKey.getWalletKeyUsingPin(pinString, walletKeyVersion)
-    const walletId = `easypid-wallet-${walletKeyVersion}`
+    const walletId = getWalletId(walletKeyVersion)
+
     const agent = new Agent({
       config: {},
       modules: {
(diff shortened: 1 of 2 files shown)


apps/easypid/src/features/pid/FunkePidSetupScreen.tsx
```diff
@@ -306,7 +309,7 @@ export function FunkePidSetupScreen() {
       if (shouldUseCloudHsm) {
         await retrieveCredential()
       }
-    } catch (error) {
+    } catch (_error) {
```
berendsliedrecht:
```suggestion
    } catch () {
```

packages/agent/src/format/formatPresentation.ts
```diff
@@ -354,7 +350,7 @@ function simplifyJsonPath(path: string, format?: ClaimFormat, filterKeys: string
     }
 
     return simplified
-  } catch (error) {
+  } catch (_error) {
```
berendsliedrecht:
```suggestion
  } catch () {
```

packages/agent/src/invitation/shareProof.ts
```diff
@@ -49,10 +50,8 @@ export const shareProof = async ({
                 entry.verifiableCredentials.find((vc) => vc.credentialRecord.id === credentialId) ??
                 entry.verifiableCredentials[0]
 
-              // Optionally use a batch credential
-              const credentialRecord = await handleBatchCredential(agent, credential.credentialRecord)
-
-              return [entry.inputDescriptorId, [{ ...credential, credentialRecord }]]
+              // NOTE: we don't support single-use credentials for PEX
```
berendsliedrecht: When will we remove support for PEX?

berendsliedrecht (approved):
All very minor feedback, just cleanup things.

PR #441, opened 2025-12-11
fix: didcomm flow

TimoGlastra:
Fixes the DIDComm flow, the issue was with the `credo-ts-didweb-anoncreds` library importing the CJS build of Credo, and it nog being able to find the registered dependencies, since the rest of the app used the ESM build 🤦

### DIFF
--- a/packages/agent/src/agent.ts
+++ b/packages/agent/src/agent.ts
@@ -77,7 +77,7 @@ export const initializeEasyPIDAgent = async ({
       x509: new X509Module({
         getTrustedCertificatesForVerification: (agentContext, { certificateChain, verification }) => {
           if (verification.type === 'credential') {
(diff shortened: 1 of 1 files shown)


packages/agent/src/agent.ts
```diff
@@ -77,7 +77,7 @@ export const initializeEasyPIDAgent = async ({
       x509: new X509Module({
         getTrustedCertificatesForVerification: (agentContext, { certificateChain, verification }) => {
           if (verification.type === 'credential') {
-            // Temprorary allow any certificates, also for PID
+            // Temporary allow any certificates, also for PID
```
hacdias:
```suggestion
            // Temporarily allow any certificates, also for PID
```

Maybe?

PR #480, opened 2026-01-28
feat: add french translations

TimoGlastra:


Wicpar:
this drifted a lot since the last time, i have pushed #522

Wicpar:
i suggest this be closed in favor of #522

PR #492, opened 2026-02-27
Feat/ts12 payments

Wicpar:
###  1. DC API flow and selection handling

  - Added support for multiple presentation/VCI handling.
  - Fixed issuance flow behavior.
  - Improved request parsing/selection deduction via dcApi request payload handling.
  - Refreshed credential providers on app foreground for better runtime consistency.
  - Vendored DC API matcher packages and updated dependency/path handling.

###  2. Credential/agent correctness

  - Fixed SD-JWT PID ordering for QES/payment scenarios.
  - Updated agent image/base64 handling to use Expo File API.

###  3. TS12 UX improvements

  - Made transaction slides scrollable.
  - Improved table row density/fallback layout behavior for long claim data.
  - Removed dev-only hardcoded TS12 test rows from tables.
  - Auto-approved TS12 payment transaction entries and skipped explicit payment review step.

###  4. Minor polish

  - Fixed OID4VCI release icon issue.


packages/agent/src/providers/W3cV2CredentialsProvider.tsx
```diff
@@ -67,12 +68,26 @@ export const W3cV2CredentialRecordProvider: React.FC<PropsWithChildren<Props>> =
...
+    setState({ w3cV2CredentialRecords, isLoading: false })
   }, [agent])
 
+  useEffect(() => {
+    void fetchRecords()
+  }, [fetchRecords])
+
+  useEffect(() => {
+    const handleAppStateChange = (nextAppState: AppStateStatus) => {
+      if (nextAppState === 'active') {
```
TimoGlastra: Why do we need this?

packages/agent/src/openid4vc/registerDcApi.ts
```diff
@@ -268,12 +264,19 @@ async function loadCachedImageAsBase64DataUrl(
 
     // In case of local image
     asset = ExpoAsset.Asset.fromModule(url)
-    return await resizeImageWithAspectRatio(logger, asset)
+    try {
+      return await resizeImageWithAspectRatio(logger, asset)
+    } catch {
+      return await readAssetAsBase64(logger, asset)
+    }
```
TimoGlastra: This catch is kinda useless right as the outer catch does the same?

packages/agent/src/openid4vc/registerDcApi.ts
```diff
@@ -232,11 +208,31 @@ async function resizeImageWithAspectRatio(logger: Logger, asset: ExpoAsset.Asset
...
+  if (lower.endsWith('.jpg') || lower.endsWith('.jpeg')) return 'jpg'
+  return 'png'
+}
+
+async function readAssetAsBase64(logger: Logger, asset: ExpoAsset.Asset): Promise<ImageDataUrl | undefined> {
+  if (!asset.localUri) return undefined
+  try {
+    const base64 = await new File(asset.localUri).base64()
+    if (!base64) return undefined
+    const mime = getImageMimeFromUri(asset.localUri)
```
TimoGlastra: Why do we load from uri instead of the file itself?

Not all images have extension

apps/easypid/src/features/share/DcApiSharingScreen.tsx
```diff
@@ -321,25 +334,7 @@ export function DcApiSharingScreenWithContext({ request }: DcApiSharingScreenPro
       ]
     }
 
-    if (entry.type === 'urn:eudi:sca:payment:1') {
-      return [
-        {
-          step: `ts12-payment-${index}`,
-          progress,
-          screen: (
-            <Ts12PaymentSlide
```
TimoGlastra: We don't use this slide anymore then?

I think in general we should restructure the flow to just show everything on a single page, and you can click to see more detail for each (qes, payments, credentials).

That's a bigger refactor, and we should make in main first.

PR #521, opened 2026-04-17
Fix biometric wallet auth and DC API Chrome compatibility

Wicpar:
## What changed

This PR centralizes wallet flow authorization around shared PIN/biometric components and helpers:

- Adds reusable `WalletPinPrompt`, `WalletFlowAuthPrompt`, and flow authorization helpers.
- Replaces duplicated PIN/auth slides and dead biometric/PIN utilities across unlock, issuance, presentation, offline, and DC API flows.
- Keeps Cloud HSM authorization PIN-only and clears the flow-scoped WSP PIN after use.
- Uses biometric unlock where available for normal wallet flow authorization instead of asking for PIN by default.
- Shows DC-API flow errors in the wallet sheet instead of immediately closing the UI, with debug details in development mode.

It also updates the SDK Digital Credentials API handling for current Chrome compatibility:

- Reads the newer Chrome request payload locations and selected credential metadata.
- Sends the response as the expected `{ protocol, data }` envelope.
- Handles `dc_api.jwt` by returning the nested `response` string shape Chrome expects.

## Why

The biometric/PIN behavior had drifted across flows. Centralizing the auth prompt and authorization policy makes one-auth-per-flow behavior easier to maintain and avoids separate implementations for OpenID, offline, and DC API presentations.

The DC API update is required because newer Chrome versions reject the older response/request structure. Without the compatibility update, Chrome will refuse wallet responses even when credential selection and signing succeed. When a DC-API request fails before a valid response can be built, the user now gets the same style of error surface as other VP flows instead of a silent close.

## Validation

- `pnpm exec biome check --unsafe apps/easypid/src/features/share/DcApiSharingScreen.tsx apps/easypid/src/features/receive/slides/InteractionErrorSlide.tsx`
- `pnpm types:check`
- `git diff --check upstream/main..HEAD`


Wicpar:
@TimoGlastra I have manually reviewed the changes as well as tested the changed flows. You can drop the second commit that just handles some errors more cleanly, as it is a bit iffy in error detection.

berendsliedrecht:
Hi @Wicpar We have added payment, and payment status, to the paradym wallet using some of the existing libraries now! Therefore we will close these pull requests as they are not needed anymore. Thanks for the work and they were a great reference while updating all the different libraries and components!

PR #522, opened 2026-04-19
Added French translations

Wicpar:
French translations have been added.


packages/translations/src/commonMessages.ts
```diff
@@ -576,6 +576,10 @@ export const supportedLanguageMessages: Record<SupportedLocale, MessageDescripto
     id: 'language.fi',
     message: 'Finnish',
   }),
+  fr: defineMessage({
+    id: 'language.fr',
+    message: 'French',
```
TimoGlastra, line 581: This doesn't have a translation in the other languages. There was a suggestion made by someone a while ago to just use the local language everywhere so you can recognize it (so "Nederlands" in all languages). Then we don't need translations, but just one for each language. 

So either add transaltions for this, or change it to not require translations for the language name and us the local name

TimoGlastra (approved):

PR #533, opened 2026-06-05
fix(sdk): id.animo.paradym deeplink working again

berendsliedrecht:

### DIFF
--- a/apps/easypid/src/app/+native-intent.tsx
+++ b/apps/easypid/src/app/+native-intent.tsx
@@ -12,6 +12,11 @@ import { credentialDataHandlerOptions } from './(app)/_layout'
 // deeplink from working on a cold startup. We updated the invitation handler to
 // be fully sync.
 export function redirectSystemPath({ path, initial }: { path: string; initial: boolean }) {
+  // short-circuit when we just want to enter the app without any link
+  if (path === 'id.animo.paradym:///') {
+    return '/'
+  }
+
   const logger = new ParadymWalletSdkConsoleLogger(LogLevel.trace)
 
   logger.debug(`Handling deeplink for path ${path}.`, {
@@ -97,14 +102,15 @@ export function redirectSystemPath({ path, initial }: { path: string; initial: b
       void Haptics.notificationAsync(Haptics.NotificationFeedbackType.Error)
       return '/'
     } catch (error) {
-      logger.info('Deeplink is not a valid invitation. Routing to home screen', {
+      logger.error('Deeplink is not a valid invitation. Routing to home screen', {
         error: error,
         message: (error as Error).message,
       })
 
       return '/'
     }
-  } catch (_error) {
+  } catch (error) {
+    logger.error('Error occurred during redirectSystemPath', { cause: error })
     return '/'
   }
 }
--- a/packages/sdk/src/ParadymWalletSdk.tsx
+++ b/packages/sdk/src/ParadymWalletSdk.tsx
@@ -440,10 +440,10 @@ function useSecureUnlockState(configuration: SetupParadymWalletSdkOptions): Secu
       },
       reinitialize,
       lock: async () => {
-        await paradym.shutdown()
-        setParadym(undefined)
         setState('locked')
+        setParadym(undefined)
         setUnlockMethod(undefined)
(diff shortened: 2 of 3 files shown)


apps/easypid/src/app/authenticate.tsx
```diff
@@ -38,12 +38,6 @@ export default function Authenticate() {
 
   const isLoading = paradym.state === 'locked' && paradym.isUnlocking
 
-  useEffect(() => {
-    if (paradym.state === 'unlocked' && redirectAfterUnlock) {
-      paradym.lock()
-    }
-  }, [])
```
TimoGlastra, line 45: This used to be important, why is it not anymore? This redirect logic is very brittle, do we know 100% sure this doesn't break anything?

apps/easypid/src/app/+native-intent.tsx
```diff
@@ -12,6 +12,11 @@ import { credentialDataHandlerOptions } from './(app)/_layout'
 // deeplink from working on a cold startup. We updated the invitation handler to
 // be fully sync.
 export function redirectSystemPath({ path, initial }: { path: string; initial: boolean }) {
+  // short-circuit when we just want to enter the app without any link
+  if (path === 'id.animo.paradym:///') {
+    return '/'
+  }
```
TimoGlastra, line 18: What does this address? I think we never just open the wallet, i think the error was in this deeplink + an invitation

berendsliedrecht: I changed so that when this function catches an error we log it. This caused an error so I handled the case where we basically open the app. With expo or any other way.
Can also remove the error log, but it can be useful to debug the deeplink stuff.

apps/easypid/src/app/authenticate.tsx
berendsliedrecht: So this was the main for the issue that https://paradym.id links didn't work anymore. It tried to resolve everything and also lock the wallet. If you test the flow before it will actually resolve the invite (qr code on the page goes to a loader) and after unlocking it will try again. I tried it with a normal deeplink and the paradym.id invite and it does work now.

TimoGlastra: Hmm okay, so it now correctly locks the wallet (and requires the pin again) when you open a deeplink? 

Because i think how the flow worked is:
1. receive deeplink -> set redirectAfterUnlock to the deeplink
2. lock the wallet (this removed code)
3. user authenticates
4. user is redirected to redirectAfterUnlock

berendsliedrecht: Ah yes the lock was removed before. I didn't think it would be a big issue due to apps not being allowed to be "alive" in the background. Fixed it now.

https://github.com/user-attachments/assets/ca23d9d2-16ab-4c5d-95d8-e87e7ac1b7da

packages/sdk/src/ParadymWalletSdk.tsx
```diff
@@ -443,10 +440,10 @@ function useSecureUnlockState(configuration: SetupParadymWalletSdkOptions): Secu
       },
       reinitialize,
       lock: async () => {
-        await paradym.shutdown()
-        setParadym(undefined)
         setState('locked')
+        setParadym(undefined)
         setUnlockMethod(undefined)
+        await paradym.shutdown()
```
berendsliedrecht, line 446: This was the fix. We had an issue where two checks were done on whether the sdk was `unlocked` which caused a race condition as the second one should start "listening" after the first one has fininshed. But this ordering removes that race condition.

apps/easypid/src/app/authenticate.tsx
TimoGlastra: What do you mean with alive? The auto lock in background was a security feature. But it could cause a eace condition so we always redirected to the lock screen and required to authenticate again. Is this still the behavior?

berendsliedrecht: Yeah, that's in the video right? Or do you mean something else?

TimoGlastra: No just wanted to check

TimoGlastra (approved):

PR #534, opened 2026-06-05
fix: allow deletion of credentials

berendsliedrecht:
Did not work due to some legacy code we had for deleting a category. Code was never used.


TimoGlastra (approved):
This also fixes the modal not closing?

berendsliedrecht:
> This also fixes the modal not closing?

Yes, it was an issue with navigating and closing the model together. It should've caused issues before as well, but I also sometimes didnt have the issue while debugging. Maybe it got more prevalent.

PR #536, opened 2026-06-05
fix: bring back trusted did entities

berendsliedrecht:


packages/sdk/src/trust/trustMechanism.ts
```diff
@@ -107,7 +108,10 @@ export const getTrustedEntities = async (
       })
       break
     case 'did':
-      trustedEntities = await getTrustedEntitiesForDid(options)
+      trustedEntities = await getTrustedEntitiesForDid({
+        ...options,
+        trustMechanismConfiguration: trustMechanismConfiguration as DidTrustMechanismConfiguration,
```
hacdias, line 113: Is this `as` necessary?

hacdias (approved):

packages/sdk/src/trust/trustMechanism.ts
berendsliedrecht: yes :( for type narrowing. Can probably be done smarter, though.

PR #542, opened 2026-06-16
feat(payments): status indication

berendsliedrecht:
## Payment status

Initial status state is `pending` and will be fetched/updated when displaying the details page. When the state is rejected/accepted it will not fetch the state anymore and keep this. UI reflects the state like before, but is now actually reflecting real data.

playground uses a 30second timeout to make the state go from pending to accepted when the amount is 100 or below, and rejected when the amount is 101 or more.


apps/easypid/src/features/wallet/components/LatestActivityCard.tsx
```diff
@@ -92,6 +116,21 @@ export function LatestActivityCard() {
...
+    if (latestActivity.type === 'payment') {
+      let description: string
+      if (['failed', 'stopped'].includes(latestActivity.status)) {
+        description = t(paymentFailed)
+      } else if (latestActivity.transactionStatus === 'RJCT') {
+        description = t(paymentRejected)
+      } else if (latestActivity.transactionStatus === 'ACSC') {
+        description = t(paymentMade)
+      } else {
+        description = t(paymentPending)
```
TimoGlastra, line 128: There's a lot more transaction codes. Are these all payment pending?

apps/easypid/src/hooks/usePaymentTransactionStatus.tsx
```diff
@@ -0,0 +1,34 @@
...
+import type { PaymentActivity } from '@paradym/wallet-sdk/storage/activityStore'
+import { useEffect } from 'react'
+
+export function usePaymentTransactionStatus(activity: PaymentActivity | undefined) {
+  const { credentials } = useCredentials()
+  const { paradym } = useParadym('unlocked')
+
+  useEffect(() => {
+    if (!activity || !paradym) return
+    if (activity.transactionStatus === 'ACSC' || activity.transactionStatus === 'RJCT') return
```
TimoGlastra, line 17: Same here, i'm not sure if we should only treat these statuses as finished and all others as not?

packages/sdk/src/openid4vc/paymentTransactionStatus.ts
```diff
@@ -0,0 +1,28 @@
...
+  try {
+    const response = await fetch(url, {
+      method: 'POST',
+      headers: {
+        Authorization: `Bearer ${token}`,
+        'Content-Type': 'application/x-www-form-urlencoded',
+      },
+      body: `transaction=${transactionHash}`,
+    })
+    const body = (await response.json()) as { status_code?: PaymentTransactionStatusCode }
```
TimoGlastra, line 24: should check response code before extracting body

packages/sdk/src/openid4vc/transaction.ts
```diff
@@ -15,6 +16,8 @@ export type FormattedTransactionDataPaymentSingle = {
     website: string
   }
   cardForTransactionId?: string
+  // hex encoded hash used for fetching the status
```
TimoGlastra: base64url

packages/translations/src/commonMessages.ts
```diff
@@ -295,6 +295,16 @@ export const commonMessages = {
     message: 'Payment failed',
     comment: 'Shown when a payment is failed',
   }),
+  paymentPending: defineMessage({
```
TimoGlastra, line 298: can you add the translation files for the new translations?

apps/easypid/src/features/activity/FunkeActivityDetailScreen.tsx
```diff
@@ -302,9 +294,6 @@ export function SharedActivityDetailSection({
                   Payment
                 </Trans>
               </Heading>
-              <Paragraph>
-                {activity.status === 'success' ? t(activityMessages.paymentMade) : t(activityMessages.paymentNotMade)}
-              </Paragraph>
```
TimoGlastra, line 307: Don't we show anything anymore on the payments status?

apps/easypid/src/hooks/usePaymentTransactionStatus.tsx
berendsliedrecht: Currently we only use 3 statuses. Since it is so tightly coupled I think it should be fine. If we want to add support for all states we need to create/find a flow diagram of all states.

apps/easypid/src/features/activity/FunkeActivityDetailScreen.tsx
berendsliedrecht: It was a duplicate field. The title already displays it very prominently, and this seemed a bit out of place.

apps/easypid/src/features/wallet/components/LatestActivityCard.tsx
berendsliedrecht: I have only implemented the states supported by the playground. Since the states are also related to the payment type we support, currently only single payments.

packages/translations/scripts/translate-missing-with-claude.js
```diff
@@ -0,0 +1,93 @@
...
+
+  const missing = readFileSync(missingPath, { encoding: 'utf-8' })
+  if (Object.keys(JSON.parse(missing)).length === 0) {
+    console.log(`Skipping ${locale}: missing.json is empty`)
+    continue
+  }
+
+  console.log(`Translating ${locale}...`)
+
+  const prompt = `You are translating Lingui catalog entries for the Paradym wallet.
```
TimoGlastra, line 71: When i run this locally it has access to the AI instructions for each language. These are important for consisentcy

berendsliedrecht:
@claude can you review this?

berendsliedrecht:
@claude could you do a rereview? The base64 issue is a mistake, it should be like that.

apps/easypid/src/crypto/WalletServiceProviderClient.ts
```diff
@@ -36,7 +63,7 @@ export const setWalletServiceProviderPin = async (pin: Array<number>, validatePi
         askar: new AskarModule({
           askar: NativeAskar.instance,
           store: {
-            id: `paradym-wallet-${walletKeyVersion}`,
+            id: getValidationStoreId(walletKeyVersion),
```
TimoGlastra, line 66: Why is there a separate store for wallet validation? This must be created when setting the pin, otherwise if you enter a pin the first time for the WSP, it will allow any value (as it creates the store the first time). And that can then lead to an issue that can not be fixed without reset.

packages/translations/scripts/extract-all-missing-translations.js
TimoGlastra, line 1: How does this relate to `packages/translations/scripts/extract-all-missing-translations.js`, can we remove that file now?

packages/translations/scripts/extract-all-missing-translations.js
```diff
@@ -0,0 +1,34 @@
...
+  console.error('Usage: node extract-all-missing-translations.js <locales-dir>')
+  process.exit(1)
+}
+
+function extractForLocale(translationPath) {
+  const resolvedTranslationPath = path.resolve(translationPath)
+  const translations = JSON.parse(readFileSync(resolvedTranslationPath, { encoding: 'utf-8' }))
+
+  const missingEntries = Object.fromEntries(
+    Object.entries(translations).filter(([, { translation }]) => translation === '')
```
TimoGlastra: If we go more into this AI translation based on missing.json, we should probably check also the the original is the same as the english translation. If the english source changes the translation won't be empty, but it should be updated. WDYT?

apps/easypid/src/crypto/WalletServiceProviderClient.ts
berendsliedrecht: > Why is there a separate store for wallet validation

That took me a while to figure out as well, not introduced in this PR. But basically what I found is that we do this to validate the pin when we do a presentation response. We compare the derived key to the actual key used to open the wallet.

packages/translations/scripts/extract-all-missing-translations.js
berendsliedrecht: Thats the same file, right?

packages/translations/scripts/extract-all-missing-translations.js
berendsliedrecht: Not entirely sure what you mean. Do you mean, we pick some english text for a field and then the `en.json` should have the same text?

packages/translations/scripts/extract-all-missing-translations.js
TimoGlastra: i mean without the `all`, the file that already existed before

packages/translations/scripts/extract-all-missing-translations.js
TimoGlastra: yeah each entry contains the source translation. which should be the same as the translation in the en.json file. Otherwise changing translations/text will never get picked up by the AI

apps/easypid/src/crypto/WalletServiceProviderClient.ts
TimoGlastra: Hmm but where do we do that, because this was using the actual wallet, not anothe wallet it seems?

And we should fix the issue with the different  pins for different stores then?

packages/translations/scripts/extract-all-missing-translations.js
berendsliedrecht: Ahhh. Yeah we never want to translate a single file, so makes sense to remove the non `all`.

packages/translations/scripts/extract-all-missing-translations.js
berendsliedrecht: Fixed it! It uses `git` under the hood to compare it. Ran a test and it seems to work.

TimoGlastra (approved):

PR #544, opened 2026-06-17
fix(sdk): setup for expo-auto-linking by using direct dependencies and not peer dependencies

berendsliedrecht:


TimoGlastra (approved):
nice

PR #545, opened 2026-06-17
feat: invalid transaction code screen

berendsliedrecht:


packages/sdk/src/openid4vc/func/acquireCredentialsPreAuthWithTransactionCode.ts
```diff
@@ -22,11 +24,22 @@ export const acquireCredentialsPreAuthWithTransactionCode = async (
...
+      txCode: options.transactionCode,
+    })
+  } catch (error) {
+    if (
+      error instanceof Oauth2ClientErrorResponseError &&
+      error.errorResponse.error === Oauth2ErrorCodes.InvalidGrant
+    ) {
+      throw new ParadymWalletInvalidTransactionCodeError()
+    }
+    throw error
```
TimoGlastra, line 41: i think they are adding a new error specifically for this case ,as invalid grant is wider. But i think for now this makes sense

TimoGlastra (approved):

TimoGlastra:
One thing to add: it should be possible to retry the tx-code maybe 1-2 times. If you make a typo, it's annoying to have to generate a new QR. it should handle it like invalid pin

berendsliedrecht:
> One thing to add: it should be possible to retry the tx-code maybe 1-2 times. If you make a typo, it's annoying to have to generate a new QR. it should handle it like invalid pin

Yes, we can reuse the same QR and try again actually. I will add some code to try it 2 additional times and then go to the error screen.

PR #549, opened 2026-06-18
expo  56 update

berendsliedrecht:
Used claude to generate the update from 54 -> 56. Worked really well as Expo has a lot of tools for AI Agents so that I just had to prompt to update it and fetch the information from expo. 

Dependent upon #542 for full testing, but app launches now!


tsconfig.json
```diff
@@ -4,6 +4,7 @@
 		"strict": true,
 		"noEmitOnError": true,
 		"baseUrl": ".",
+		"ignoreDeprecations": "6.0",
```
TimoGlastra: Which deprecations? Are they easy to fix?

TimoGlastra (approved):

tsconfig.json
berendsliedrecht: Let me checkkkkk

berendsliedrecht: Well we just had to remove `baseUrl` and thats it. Nice!

PR #552, opened 2026-06-24
chore: convert the current CLAUDE.md to a translation skill

berendsliedrecht:


.claude/skills/translations/SKILL.md
```diff
@@ -0,0 +1,72 @@
...
+
+```bash
+pnpm translations:ai
+```
+
+This pipeline runs for every locale under `apps/easypid/src/locales` except `en`:
+
+1. `translations:extract` in `apps/easypid` — refresh catalogs from source.
+2. `extract-all-missing-translations` — write a `missing.json` next to each locale's `messages.json`.
+3. `translate-missing-with-claude` — invoke the `claude` CLI per locale, using each locale's `AI_INSTRUCTIONS.MD` and the `missing.json`. This step is what loads this skill.
```
TimoGlastra: this feels a bit meta. Claude will now use `claude` to create translations. Is it not better to learn claude directly to do what is in the `translate-missing-with-claude` instead, so that it can run it itself, or start a subagent?

TimoGlastra: Even more meta, the `translate-missing-with-claude` points back at this file again

berendsliedrecht: Yeah good point 😅 I will see if it can be improved a bit.

berendsliedrecht: Wait no its not really meta. the `pnpm translations:ai` will not be invoked by ai. But instead of passing `claude.md` to the translate-missing-with-claude it uses the `SKILL.md` now. I think keeping the context of the scripts is good to give to the AI, but it will not spawn child AI's to do the translations as we only run `pnpm translations:ai`.

I will change it a bit to add that `pnpm translations:ai` should not be started by claude, but it can be. And then we have baby claudes, but IMO thats not so bad.

berendsliedrecht: I changed it so that you sytart the translations now from claude. I would assume a different AI agent would also work as it does not rely on claude specific APIs.

package.json
```diff
@@ -12,7 +12,8 @@
 		"style:fix": "biome check --write --unsafe",
 		"release:sdk": "pnpm build && pnpm changeset publish --no-git-tag",
 		"changeset-version:sdk": "pnpm changeset version && pnpm style:fix",
-		"translations:ai": "node packages/translations/scripts/translate-all.js"
+		"translations:ai:prepare": "node packages/translations/scripts/translate-prepare.js",
```
berendsliedrecht, line 15: basically flow is:

ai runs translations:ai:prepare
then ai trabslates all the missing.json
and then ai runs translations:ai:finalize.

hacdias (approved):

PR #556, opened 2026-06-26
chore(deps): update dependencies

berendsliedrecht:
- **chore(deps): update dependencies**
- **chore: update dependencies**


biome.json
```diff
@@ -24,7 +24,6 @@
 	"linter": {
 		"enabled": true,
 		"rules": {
-			"recommended": true,
```
TimoGlastra, line 27: should add `preset: recommended` i think?

TimoGlastra (approved):

TimoGlastra:
Did you test in both android and ios?

biome.json
berendsliedrecht: Thats the default

PR #558, opened 2026-07-01
fix(ui): only show disclosed metadata in the UI

berendsliedrecht:

### DIFF
--- a/apps/easypid/src/features/receive/FunkeOpenIdCredentialNotificationScreen.tsx
+++ b/apps/easypid/src/features/receive/FunkeOpenIdCredentialNotificationScreen.tsx
@@ -13,7 +13,7 @@ import {
   useParadym,
 } from '@paradym/wallet-sdk'
 import { useLocalSearchParams } from 'expo-router'
-import { useCallback, useEffect, useState } from 'react'
+import { useCallback, useEffect, useRef, useState } from 'react'
 import { setWalletServiceProviderPin } from '../../crypto/WalletServiceProviderClient'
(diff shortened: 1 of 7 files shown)


apps/easypid/src/features/receive/FunkeOpenIdCredentialNotificationScreen.tsx
```diff
@@ -71,7 +71,13 @@ export function FunkeCredentialNotificationScreen() {
     [isDevelopmentModeEnabled]
   )
 
+  // The authorization challenge / presentation request_uri the issuer hands us during
+  // resolve is one-shot on the server side, so re-firing this effect would 404 the second call.
+  const resolvingOfferUriRef = useRef<string | undefined>(undefined)
```
berendsliedrecht, line 76: I think this got lost in a different PR, but it is required so the flow does not crash due to a double request

apps/easypid/src/features/share/FunkeOpenIdPresentationNotificationScreen.tsx
```diff
@@ -62,8 +62,12 @@ export function FunkeOpenIdPresentationNotificationScreen() {
     comment: 'Shown when authentication with PIN fails for other reasons',
   })
 
+  // request_uri in OpenID4VP is one-shot on the verifier side, so we must guarantee a single
+  // resolve per uri across re-renders and React's dev-mode double-invoked effects.
+  const resolvingUriRef = useRef<string | undefined>(undefined)
```
berendsliedrecht, line 67: I think this got lost in a different PR, but it is required so the flow does not crash due to a double request

hacdias (approved):
Gucci.

PR #560, opened 2026-07-07
feat: add did:webvh support

IliasElbarnoussi:
Register `WebVhModule`, `WebVhDidResolver`, and `WebVhAnonCredsRegistry` on the agent so the wallet can resolve `did:webvh` DIDs and verify credentials and presentations from `did:webvh` issuers and verifiers.

`DidsModule` now passes an explicit resolver list (`web`, `key`, `peer`, `jwk`, and `webvh`), since providing resolvers replaces the default resolver configuration.

Signed-off-by: Ilias Elbarnoussi <ilias@animo.id>

### DIFF
--- a/packages/sdk/package.json
+++ b/packages/sdk/package.json
@@ -35,6 +35,7 @@
 		"@credo-ts/didcomm": "catalog:",
 		"@credo-ts/openid4vc": "catalog:",
 		"@credo-ts/react-native": "catalog:",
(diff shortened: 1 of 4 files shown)


berendsliedrecht (approved):

berendsliedrecht:
I think you need to regenerate the pnpm-lock.yaml file and commit it.

PR #576, opened 2026-08-25
fix(sdk): don't append wallet key version to store id twice

berendsliedrecht:
The unlock flow appended the wallet key version to the store id and `setupAgent` appended it again, so stores ended up as `easypid-wallet-1-1` instead of `easypid-wallet-1`. The id is now versioned only in `getBaseModules`.

Existing stores with the doubled id (dev only, no production installs) are migrated on unlock by renaming the wallet directory. The rename is atomic and best effort; remove after a few releases.

Verified on device: legacy store is renamed and opens with existing data intact, fresh installs unaffected.

### DIFF
--- a/packages/sdk/src/ParadymWalletSdk.tsx
+++ b/packages/sdk/src/ParadymWalletSdk.tsx
@@ -1,7 +1,9 @@
 import { AskarStoreInvalidKeyError } from '@credo-ts/askar'
 import { CredoError, type X509ModuleConfigOptions } from '@credo-ts/core'
+import { agentDependencies } from '@credo-ts/react-native'
 import { QueryClient, QueryClientProvider, useQuery } from '@tanstack/react-query'
 import { createContext, type PropsWithChildren, useContext, useState } from 'react'
+import { moveFile } from 'react-native-fs'
 import {
   type AgentForAgentType,
   type AgentType,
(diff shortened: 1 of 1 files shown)


TimoGlastra (approved):

packages/sdk/src/ParadymWalletSdk.tsx
```diff
@@ -381,9 +382,19 @@ function useSecureUnlockState(configuration: SetupParadymWalletSdkOptions): Secu
...
+          // Dev-only migration from the double-versioned store id (e.g. paradym-wallet-1-1).
+          // No production installs have the old id. Remove after a few releases.
+          const fs = new agentDependencies.FileSystem()
+          const walletDirectory = `${fs.dataPath}/wallet/${pws.walletId}`
+          const legacyWalletDirectory = `${walletDirectory}-${walletKeyVersion}`
+
+          if (!(await fs.exists(walletDirectory)) && (await fs.exists(legacyWalletDirectory))) {
+            await moveFile(legacyWalletDirectory, walletDirectory).catch(() => {})
+          }
+
```
TimoGlastra, line 397: I think we don't really need this for dev. It's complexity and doesn't affect production

berendsliedrecht: I added it so that apps in TestFlight will still work. Right now sprind and all test users have to reinstall the app.

PR #577, opened 2026-08-29
feat: add iso mdoc dc api

TimoGlastra:
Adds support for the DC API on mdoc. see notes in https://github.com/openwallet-foundation/credo-ts/pull/2922 and https://github.com/animo/expo-digital-credentials-api/pull/8 for details.

Also makes numerous improvements to the SDK, refactors a lot for it to work well with the DC API, and fixes several bugs. Some changes:
- show missing attributes as missing
- fix anoncreds attribute rendering (broken in main)
- treat mdoc requests with multiple doc requests as OR (proximity treated it as AND).
- Extract a lot of translations into common messages to reduce duplication
- Make sure the SDK does not depend on translations or any other packages
- Rework the activity store, it was causing significant lag on my device due to passing full payloads encoded in paths, (multiple mbs with images). The store now has a v3, only stores the paths, and resolves the credential dynamically if it's still on the device. Otherwise you just see the paths you shared. All previous activities are migrated to v3 store. The v3 store also splits activities into separate records, so we don't have to load all activities at once.
- Fix several performance issues, causing a lot of credential for display calculations to run a lot of times.

https://github.com/user-attachments/assets/2a8488e9-ceee-4c2e-b977-60eeb8e1fef5


apps/wallet/src/config/paradym.ts
```diff
@@ -1,13 +1,30 @@
+import { walletId } from '@app/config/wallet'
 import { eudiTrustList, trustedDidEntities, trustedOpenId4VciIssuerEntities, trustedX509Entities } from '@app/constants'
+import { mmkv } from '@app/storage/mmkv'
+import { resolveDcApiDisplay } from '@app/utils/dcApiRegisterOptions'
+import { resolveAttributeLabel } from '@app/utils/resolveAttributeLabel'
 import type { SetupParadymWalletSdkOptions } from '@paradym/wallet-sdk'
-import { LogLevel } from '@paradym/wallet-sdk'
+// The leaf rather than the barrel: the credential request UI imports this configuration too, and
+// the barrel would pull the entire SDK into its bundle.
+import { LogLevel } from '@paradym/wallet-sdk/logging/ParadymWalletSdkLogger'
```
berendsliedrecht, line 9: Does the wallet not import the entire paradym sdk anywhere else anyways?

apps/wallet/src/features/dc-api/components/DcApiRequestedCards.tsx
```diff
@@ -0,0 +1,376 @@
...
+  const unavailable = unsatisfied.filter((entry) => entry.partialMatches.length === 0)
+
+  if (submission.entries.length === 0) {
+    return <Paragraph color="$danger-500">{t(messages.nothingRequested)}</Paragraph>
+  }
+
+  const unavailableHeading = t(commonMessages.unavailableCardsHeading)
+  const unmetAttributeMessages = getUnmetAttributeMessages(submission)
+
+  return (
```
berendsliedrecht, line 55: Would it make more sense to have just separate components for each condition? This seems like a pain to update for a specific case. If it results in 10 components for example it might not be the best idea

apps/wallet/src/features/dc-api/components/DcApiRequestedCards.tsx
```diff
@@ -0,0 +1,376 @@
...
+  backgroundImage,
+  issuerImage,
+  textColor,
+  attributeNames,
+  missingAttributeNames,
+  isExpanded = false,
+  onToggle,
+  children,
+}: CardProps) {
+  const isMissing = (name: string) => missingAttributeNames?.includes(name) ?? false
```
berendsliedrecht, line 211: Maybe define this outside of the component as well, like with `toRows`

apps/wallet/src/features/dc-api/components/DcApiVerifierSection.tsx
```diff
@@ -0,0 +1,134 @@
...
+    .filter((entity) => entity.entityId !== verifier.entityId)
+    .map((entity) => ({ ...entity, demo: isDemoTrustedEntity ? true : entity.demo }))
+
+  const isRecognized = approvedBy.length > 0 || entityIsTrustAnchor
+
+  return (
+    <YStack gap="$4">
+      <YStack gap="$3" ai="center">
+        {/* `$white`, not `$background`: the logo is the verifier's artwork, commonly dark on
+            transparent, so this is a plate under it rather than a surface of ours to darken. */}
```
berendsliedrecht, line 41: Maybe we can check it dynamically to see which background to display? Maybe there is a small efficient algorithm to detect the median color? I can imagine some companies have a white/light logo as well :)

apps/wallet/src/features/dc-api/DcApiScreen.tsx
```diff
@@ -0,0 +1,405 @@
+import type { DcApiRequest } from '@animo-id/expo-digital-credentials-api/request-handler'
+// Deep import: `@app/hooks` pulls in the over-asking AI and the payment hooks, none of which this
```
berendsliedrecht, line 2: We should remove the over-asking AI :).

apps/wallet/src/features/dc-api/index.tsx
```diff
@@ -0,0 +1,69 @@
+// Entry point of the credential request UI on both platforms. It is bundled separately from the
```
berendsliedrecht, line 1: Should we make sure this is never imported in the apps/app itself? Or does that not matter?

apps/wallet/src/features/dc-api/index.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+/**
+ * The same catalogs the app registers, each behind a thunk.
+ *
+ * Metro has no code splitting, so all seven are in this bundle either way — but a compiled catalog
+ * is one large object literal, and only the one that will actually be read has to be *evaluated*
+ * and kept on the heap. The provider extension is killed at a hard memory limit, so six catalogs
+ * this request will never show are worth not parsing.
+ */
+const catalogs: Record<SupportedLocale, () => { messages: Messages }> = {
```
berendsliedrecht, line 32: I guess that defeats the purpose of this, but we cannot have a shared method for this? To avoid getting out-of-sync

apps/wallet/src/features/didcomm/InboxScreen.tsx
```diff
@@ -1,4 +1,4 @@
-import { Trans, useLingui } from '@lingui/react/macro'
+import { useLingui } from '@lingui/react/macro'
```
berendsliedrecht, line 1: Should we maybe just let the translation package export this?

My autocomplete can never find `useLingui` and it would be nice to have the package/translation provide everything for translation, and not purely rely on a transitive dep.

apps/wallet/src/features/notifications/NotificationErrorBoundary.tsx
```diff
@@ -0,0 +1,74 @@
...
+}
+
+/**
+ * The screen recovers, so without this the error would leave no trace at all.
+ *
+ * Console only, not the SDK logger: the boundary has to hold wherever it is placed, including where
+ * the SDK is not provided or the wallet is not unlocked, so it depends on nothing the SDK provides.
+ */
+function logRenderError(error: Error) {
+  console.error('Unhandled error while rendering a request', error)
```
berendsliedrecht, line 25: You should be able to import `ParadymWalletSdkConsoleLogger` that does not rely on any SDK functionality, but it does also add it to the logging trace

berendsliedrecht:
@claude review this PR.

apps/wallet/src/features/wallet/OfflineQrScreen.tsx
```diff
@@ -44,19 +59,34 @@ export function OfflineQrScreen() {
...
+    if (!arePermissionsGranted) {
       setQrCodeData(undefined)
+      return
     }
-  }, [arePermissionsGranted])
+
+    void getMdocQrCode()
+      .then(setQrCodeData)
+      .catch((error) => {
+        // NOTE: iOS automatically handles Bluetooth permissions, so we can't easily detect if
```
berendsliedrecht, line 70: I hate this....

apps/wallet/src/hooks/useSyncSdkLocale.tsx
```diff
@@ -0,0 +1,17 @@
+import { useLocale } from '@package/translations'
+import { useParadym } from '@paradym/wallet-sdk'
+
+/**
+ * Keeps the wallet rendering credentials in the language the app is showing.
+ *
+ * The locale reaches the SDK from here rather than the SDK reading the app's i18n, so nothing it
+ * publishes depends on the wallet's private packages.
+ */
+export function useSyncSdkLocale() {
```
berendsliedrecht, line 10: Nice!

apps/wallet/src/storage/mmkv.ts
```diff
@@ -1,3 +1,5 @@
-import { createMMKV } from 'react-native-mmkv'
+import { getSharedMmkv } from '@paradym/wallet-sdk/storage/sharedMmkv'
```
berendsliedrecht, line 1: Should we export this from the root maybe?

apps/wallet/src/utils/sharedPidSetup.ts
```diff
@@ -8,6 +8,7 @@ import { OnboardingIdCardVerify } from '@app/features/onboarding/screens/id-card
 import type { MessageDescriptor } from '@lingui/core'
```
berendsliedrecht, line 8: `pidSetup` is not relevant anymore, right?

apps/wallet/package.json
```diff
@@ -59,6 +62,8 @@
 		"react-native-gesture-handler": "catalog:",
 		"react-native-nitro-modules": "catalog:",
 		"react-native-qrcode-svg": "catalog:",
+		"react-native-quick-base64": "catalog:",
```
berendsliedrecht, line 65: why are these added now?

apps/wallet/package.json
```diff
@@ -80,7 +85,7 @@
 				"listUnknownPackages": false,
 				"exclude": [
 					"react-native-fs",
-					"@animo-id/expo-mdoc-data-transfer",
+					"expo-mdoc-data-transfer",
```
berendsliedrecht, line 88: Did this change?

apps/wallet/tamagui.config.ts
```diff
@@ -39,6 +39,83 @@ const tokens = createTokens({
   },
 })
 
+/**
```
berendsliedrecht, line 42: We have a darktheme wallet now as well?! Woah can you share some screenshots?

packages/app/src/components/CardWithAttributes.tsx
```diff
@@ -23,14 +22,29 @@ interface CardWithAttributesProps {
...
+   * attributes when both are given.
+   */
+  disclosedPaths?: FormattedSubmissionEntrySatisfiedCredential['disclosed']['paths']
   isExpired?: boolean
   isRevoked?: boolean
   isNotYetActive?: boolean
 }
 
+/** Two attributes to a row. */
+function toRows(attributes: string[]) {
```
berendsliedrecht, line 40: Maybe this should be util, I have seen it before.

packages/app/src/components/CardWithAttributes.tsx
```diff
@@ -39,50 +53,42 @@ export function CardWithAttributes({
...
 
+  // What the card holds first, then what it lacks, each group starting on a row of its own.
   const groupedAttributes = useMemo(() => {
-    const result: Array<[string, string | undefined]> = []
-    for (let i = 0; i < formattedDisclosedAttributes.length; i += 2) {
-      result.push([formattedDisclosedAttributes[i], formattedDisclosedAttributes[i + 1]])
-    }
-    return result
-  }, [formattedDisclosedAttributes])
+    const isMissing = (attribute: string) => missingAttributes?.includes(attribute) ?? false
```
berendsliedrecht, line 67: This whole code.

Can we reuse it? Or does it have to do with the DcApi bundle?

packages/sdk/src/config/index.ts
```diff
@@ -0,0 +1,175 @@
...
+import type { TrustMechanismConfiguration } from '../trust/trustMechanism'
+import type { ResolveAttributeLabel } from './attributeLabel'
+import type { ResolveDcApiDisplay } from './dcApiDisplay'
+
+/**
+ *
+ * Base id of the askar store, used when the wallet does not provide one
+ *
+ */
+export const defaultWalletId = 'paradym-wallet'
```
berendsliedrecht, line 12: Is this still relevant for our app? Or do we always rename it to `easypid-wallet`? JUst curious.

packages/sdk/src/config/locale.ts
```diff
@@ -0,0 +1,53 @@
...
+ *
+ * Safe to call while rendering, so the components rendered after the caller already read the new
+ * locale. Subscribers are told in a microtask for that reason: updating another component while one
+ * renders is not allowed. Those that rendered with the new locale in the meantime do not render again.
+ */
+export function setLocale(locale: string): void {
+  if (locale === currentLocale) return
+
+  currentLocale = locale
+  queueMicrotask(() => {
```
berendsliedrecht, line 36: Whats that?

packages/sdk/src/dcApi/appGroup.ts
```diff
@@ -0,0 +1,37 @@
...
+  } catch {
+    return undefined
+  }
+}
+
+/**
+ * Same layout Credo's `ReactNativeFileSystem` uses, so only the base path differs.
+ */
+export function getAppGroupStoreDirectory(containerPath: string, storeId: string) {
+  return `${containerPath}/.afj/wallet/${storeId}`
```
berendsliedrecht, line 28: DO we want to change `.afj` in the credo 0.8 or 0.9 release?

packages/sdk/src/format/presentationExchangeRequest.ts
```diff
@@ -155,6 +158,9 @@ export function formatDifPexCredentialsForRequest(
         description: requirement.purpose,
         isSatisfied: false,
         requestedAttributePaths: requestedAttributePaths,
+        // Credo only reports the credentials that satisfy an input descriptor, not the ones that
+        // come close, so there is nothing to show missing attributes for.
+        partialMatches: [],
```
berendsliedrecht, line 163: Do we want to add this feature to credo?

packages/sdk/src/kms/RsaVerificationKeyManagementService.ts
```diff
@@ -0,0 +1,130 @@
+import type { AgentContext } from '@credo-ts/core'
+import { Kms } from '@credo-ts/core'
+import { constants, createPublicKey, verify } from 'react-native-quick-crypto'
```
berendsliedrecht, line 3: Ahh I see now why this is imported.

packages/sdk/src/providers/SdJwtVcProvider.tsx
```diff
@@ -68,29 +68,30 @@ export const SdJwtVcRecordProvider: React.FC<PropsWithChildren<Props>> = ({ agen
 
   useEffect(() => {
     void agent.sdJwtVc.getAll().then((sdJwtVcRecords) => setState({ sdJwtVcRecords, isLoading: false }))
-  }, [agent.sdJwtVc.getAll])
+  }, [agent])
 
+  // Only the agent: re-running this on every state change tore down and rebuilt all three
+  // subscriptions on every record event, and the handlers closed over the state they were created
```
berendsliedrecht, line 74: I think I checked this with claude before for the performance and it seemed in the end that the statement it made was not true, therefore this fix did not improve the performance. Did you double check it? Maybe it is a slightly different fix then I had.

packages/sdk/src/utils/format.ts
```diff
@@ -0,0 +1,126 @@
+import { getLocale } from '../config/locale'
+
+const getLocaleForFormat = () => getLocale()
```
berendsliedrecht, line 3: Seems a bit redundant

berendsliedrecht (approved):
Very nice!!!! I think no comments are blocking just a bunch of questions and minor things.

apps/wallet/src/config/paradym.ts
TimoGlastra: No. The DC API entrypoint is separate in the SDK and it does not pull in the whole sdk. That's on purpose, on iOS memory is limited, we avoid importing several heavy deps we use (e.g. react-native-skia) that are not needed in the DC API variant.

apps/wallet/src/features/dc-api/components/DcApiRequestedCards.tsx
TimoGlastra: The conditions can combine in the UI, so it would only be for the header and description. 

I can extract the header title and description extraction into a function? But separate will be difficult, as the header just tries to match what the view shows (it shows all matched cards, cards that are missing, cards that are matched but are missing attributes, cars that are matched and have the attribute, but value does not match).

apps/wallet/src/features/dc-api/index.tsx
TimoGlastra: I can look if there's a biome rule that allows to restrict that. I think the other way around matters more (dc-api importing main app entrypoint)

apps/wallet/src/features/dc-api/index.tsx
TimoGlastra: Yeah we can reuse, this just optimizes by dynamically loading, but we can do that in the main app as well.

apps/wallet/src/features/didcomm/InboxScreen.tsx
TimoGlastra: Then the Macro doesn't work 😭 

See the [translation package readme](https://github.com/animo/paradym-wallet/blob/2d68168863dd8e78f883b752f420dc46aa2d7108/packages/translations/README.md#defining-translatable-components):
> Due to how macros work with the Bable compiler, the macros need to be imported from the lingui package directly.

apps/wallet/src/storage/mmkv.ts
TimoGlastra: MMKV is used by the DC API entrypoint, so it's a sub-import on purpose, to not pull in the whole SDK...

apps/wallet/src/utils/sharedPidSetup.ts
TimoGlastra: not really, but we still have the code in the wallet. We can maybe do some cleanup once this PR is merged ? Remove ai-oversasking, pid-setup etc..?

apps/wallet/package.json
TimoGlastra: react-native-quick-crypto depends on it, which we added to support RSA verification.

apps/wallet/package.json
TimoGlastra: It already changed a while ago when it was moved to OWF, we never updated the package in the wallet

apps/wallet/tamagui.config.ts
TimoGlastra: No, it's just for the DC API screen. iOS does not correctly take the theme for the DC API activity, so even if you force light (like we do for the wallet) it will still just look at your device theme for the header etc.. So i made the DC API screen dark mode capable to work around it.

IMG_4345

packages/sdk/src/config/index.ts
TimoGlastra: for our app we use `easypid-wallet`, but SDK users will by default use `paradym-wallet`

packages/sdk/src/config/locale.ts
TimoGlastra: Basically defer running the function until the end of the loop, but before the next loop. I think it helps with mid-render re-renders (although I'd have to dig into why claude added it again to be sure)

packages/sdk/src/dcApi/appGroup.ts
TimoGlastra: we looked into it, and decided to not do it, due to the complexity of migration / checking different paths etc. We can do it, but would have to make sure it does not cause any issues

packages/sdk/src/format/presentationExchangeRequest.ts
TimoGlastra: I'm not sure we want to invest any resources into PEX. Basically for DCQL it already does this, and PEX we just show the no matches screen, not this creential matches but misses this attribute. I think that's fine as we will drop legacy Openid4vc soon-ish

packages/sdk/src/providers/SdJwtVcProvider.tsx
TimoGlastra: I will have to check. The performance was bad, and this was one of a few things claude did, and it improved performance a lot. I think in the end the activities were the biggest contirbutor to bad performance

packages/sdk/src/config/locale.ts
TimoGlastra: looked into it, it's because this is called during rendering, and we shouldn't update external stores (`useSyncExternalStore`) during render.

packages/sdk/src/providers/SdJwtVcProvider.tsx
TimoGlastra: yeah  i think you're right

apps/wallet/src/features/dc-api/components/DcApiVerifierSection.tsx
TimoGlastra: it is possible, but requires some engineering, and also means we have to pull skia into the DC API bundle. I'd like to do it as a follow up if we want it

apps/wallet/src/features/dc-api/index.tsx
TimoGlastra: this is quite hard to do with bimoe, I'm adding a script that checks the bundle size of both. This way it's easy to notice if the DC API suddenly grows by a lot (i had that, and it was causing issues)

